PR #41, opened 2021-09-04
Sinowealth based Daly BMS support

Carstijn:
This implements support for Sinowealth chipset based BMS, for now limited to SH367303 / 367305 / 367306 / 39F003 / 39F004 / BMS_10 chips with up to 10 cells. I have tested this with my Daly SmartBMS 4S 250A BMS, this is based on the SH39F003 chipset. My battery bank is a lishen 4s3p, 12 cell 840 Ah 12V bank installed on my boat. Using a solar MPPT and Multiplus 12/3000 for charging and Venus GX device + BMV712 for controlling the setup.

**What works:**
- Reading SOC, Voltage and Current
- Reading Cell Voltages and passing Min and Max values
- Reading capacity and capacity remaining
- Reading BMS status, charge/discharge fets (not all cases tested)
- Reading Alarm flags (not all cases tested)
- Reading temperature

**What kinda works**
- Reading cycles
Reading cycles is implemented, but does not match bluetooth app value

**What is not implemented**
The BMS has more data avaiable in the app, I am not able to read this data atm. It seems there is some way to read  "extended parameter" pages, no clue on how to do this. Because of this some values are hard coded, like cell voltage thresholds and discharge and charge current limits.

**On this pull request**
- Python is not my first language, so codestyle and code improvemt suggestions are welcome.
- It is only tested on my setup, in limited fasion, let me know what works and does not work in your case
- My vacation ends tomorrow, so I have less/no access to the setup on a daily basis and have to see when I am back at my boat to test new changes.
- I am not sure on what needs to be done to complete this pullrequest, so @Louisvdw please fill me in on what is missing.

### DIFF
--- a/etc/dbus-serialbattery/sinowealth.py
+++ b/etc/dbus-serialbattery/sinowealth.py
@@ -215,43 +215,42 @@ class Sinowealth(Battery):
         return cell_voltage[0]/1000
 
     def read_temperature_data(self):
-        # TODO: get number of supported sensors and select correct one to read,
-        # currently none of the temperatures match the one in the bluetooth app
+        if self.temp_sensors is None:
+            self.read_pack_config_data()
         temp_ext1_data = self.read_serial_data_sinowealth(self.command_temp_ext1)
-        # check if connection success
         if temp_ext1_data is False:
             return False
-            
+        kelvin = 273.1
         temp_ext1 = unpack_from('>H', temp_ext1_data[:-1])
-        logger.info(">>> INFO: BMS external temperature 1: %f C", temp_ext1[0]/100 )
+        logger.info(">>> INFO: BMS external temperature 1: %f C", temp_ext1[0]/10 - kelvin )
 
         self.temp1 = temp_ext1[0]/100
         
-        temp_ext2_data = self.read_serial_data_sinowealth(self.command_temp_ext2)
-        # check if connection success
-        if temp_ext2_data is False:
-            return False
+        if self.temp_sensors is 2:
+            temp_ext2_data = self.read_serial_data_sinowealth(self.command_temp_ext2)
+            if temp_ext2_data is False:
+                return False
             
-        temp_ext2 = unpack_from('>H', temp_ext2_data[:-1])
-        logger.info(">>> INFO: BMS external temperature 2: %f C", temp_ext2[0]/100 )
+            temp_ext2 = unpack_from('>H', temp_ext2_data[:-1])
+            logger.info(">>> INFO: BMS external temperature 2: %f C", temp_ext2[0]/10 - kelvin )
 
-        self.temp2 = temp_ext2[0]/100
+            self.temp2 = temp_ext2[0]/100
         
+        # Internal temperature 1 seems to give a logical value 
         temp_int1_data = self.read_serial_data_sinowealth(self.command_temp_int1)
-        # check if connection success
         if temp_int1_data is False:
             return False
             
         temp_int1 = unpack_from('>H', temp_int1_data[:-1])
-        logger.info(">>> INFO: BMS internal temperature 1: %f C", temp_int1[0]/100 )
+        logger.info(">>> INFO: BMS internal temperature 1: %f C", temp_int1[0]/10 - kelvin )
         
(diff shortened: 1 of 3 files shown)


etc/dbus-serialbattery/utils.py
```diff
@@ -58,7 +58,7 @@ def read_serial_data(command, port, baud, length_pos, length_check, length_fixed
             count = 0
             data = bytearray(res)
             while len(data) <= length + length_check:
-                res = ser.read(length + length_check)
+                res = ser.read((length + length_check) - len(data))
```
Louisvdw, line 61: Can you give me more information on why this change was needed as it could influent the existing BMS that use this function. If you give it more bytes to read it, read will still only read what is available in the buffer. So more is good. 
Did you get any problems in the Sinowealth based Daly using the read without the  -len(data) ?

etc/dbus-serialbattery/sinowealth.py
```diff
@@ -0,0 +1,266 @@
...
+        
+        self.cell_max_no = max(self.cell_voltages, key=int) - 1
+        self.cell_max_voltage = max(self.cell_voltages.values())
+        self.cell_min_no = min(self.cell_voltages, key=int) - 1
+        self.cell_min_voltage = min(self.cell_voltages.values())
+        
+        logger.info(">>> INFO: Max cell voltage: %u:%f", self.cell_max_no, self.cell_max_voltage)
+        logger.info(">>> INFO: Min cell voltage: %u:%f", self.cell_min_no, self.cell_min_voltage)
+        
+        self.max_battery_voltage = 3.65 * self.cell_count
```
Louisvdw, line 204: You already set this in get_settings() function and it does not change. Not sure why it is here in cell data as well?

etc/dbus-serialbattery/sinowealth.py
```diff
@@ -0,0 +1,266 @@
...
+          self.read_pack_config_data()
+          
+        cell_index = 1
+        while cell_index <= self.cell_count:
+          self.cell_voltages[cell_index] = self.read_cell_voltage(cell_index)
+          if self.cell_voltages[cell_index] is False:
+            return False
+          cell_index += 1
+        
+        self.cell_max_no = max(self.cell_voltages, key=int) - 1
```
Louisvdw, line 196: If you use the cells object that exist in battery.py instead of your own cell_voltages then the min/max values will be done for you in battery as well.

etc/dbus-serialbattery/sinowealth.py
```diff
@@ -0,0 +1,266 @@
...
+        temp_sens_mask = int(~(1 << 6))
+        self.temp_sensors = 1 if (pack_config_data[1] & temp_sens_mask) else 2 # one means two
+        logger.info(">>> INFO: Number of temperatur sensors: %u", self.temp_sensors)
+        return True
+        
+    def read_cell_data(self):
+        if self.cell_count is None:
+          self.read_pack_config_data()
+          
+        cell_index = 1
```
Louisvdw, line 189: Rather use the existing cell object from battery.py for the cell data

Louisvdw (changes requested):

Carstijn:
@Louisvdw thanks for the review, valid point! I will make the changes and try to test them before posting again.

Louisvdw:
@Carstijn 
Here is an example of how your read_cell_data() function can be changed to use the Cell object (for some reason I get a github error page trying to create a pull request on your file/repo.

```
def read_cell_data(self):
	if self.cell_count is None:
	  self.read_pack_config_data()
	# clear the list
	for c in self.cells:
		self.cells.remove(c)
	# get up to the first 16 cells
	for c in range(self.cell_count):
		cell = Cell(False)
		cell.voltage = self.read_cell_voltage(c+1)
		if self.cell_voltages[cell_index] is False:
			return False
		self.cells.append(cell)
	return True
```

Carstijn:
@Louisvdw I made some changes, to fix your comments, pleas re-review.

About the list, creating and deleting it does not feel very optimal, do you have something about my latest solution? (creating once in settings and only updating)

Louisvdw (approved):
Great work adding the Sinowealth Daly

PR #72, opened 2021-11-15
Online status

### LINKED ISSUE #17: Show when bank in online/offline
Show when bank in online/offline.
If connection is lost this bank online/offline count can be used to indicate this. 
Also make sure all value updates can handle handle it if no value is retrieved.

Louisvdw:
Closes #17 to show when bank is offline when no reply is recieved in 10 iterations (10 sec. for most batteries)
Exit with status 1 for python 3 when no battery is found.
Fix on device instance


nhybrow:
Hi,
How to use INVERT_CURRENT_MEASUREMENT flag ?
I put it in the utils.py file, with = -1 argument, but doesn't work
thanks in advance

PR #75, opened 2021-11-17
Renogy initial beta

gambleben:
First attempt to get Renogy Smart Lithium battery working with this driver.

Implemented code to read battery serial number, voltage, current, and state of charge.


Louisvdw (commented):
What is the purpose of the etc/dbus-serialbattery/Renogy Test Script.py file. 
Is it something other users will also use, or is it just a test file?

gambleben:
I had created the Renogy Test Script.py just to allow me to do some testing from my PC.  I removed it from the latest commit.

Louisvdw (approved):

Issue #13, opened 2021-03-15
Add bluetooth support


Louisvdw:
It would be good to have the driver use bluetooth to connect from the VenusOS device to the BMS
The latest rPi images has bluetooth enabled. A GX device will need a USB bluetooth dongle.

It will need to

- [ ]  Pair the device and the BMS
- [ ] Send commands to connected bluetooth
- [ ] Read values from connected bluetooth

damonblumenstein:
I would love to see this come to fruition. I have 2 propane sensors on BT now and its awesome. This would make total sense!

TheEldar:
did some tests with a Raspi and my Renogy RBT200LFP12-BT. Sorry it's not for the ANT BMS. But maybe it will also Help to implement the BT Support.

The Solar Monitor from Olen works and connects to the BT module of the Renogy. The scripts are meant for MQTT but maybe you can use the BT part and for integration in the GX devices?
[Olen on GitHub](https://github.com/Olen/solar-monitor)
[](url)
And an other Script on reddit
[Link to Reddit Post](https://www.reddit.com/r/SolarDIY/comments/frafmv/comment/hoqc2wu/?utm_term=38500721598&context=3&utm_medium=comment_embed&utm_source=embed&utm_name=a9879548-ce16-11ec-9708-ae6ecb37ff8a)

baranator:
Last week i got the last parts for my victron battery system and tried to wire it up. the dbus-serialbattery wouldn't work. After some testing i came to the conclusion out that the serialport of my jk-b2a24s20p is broken as i couldn't get it running with any serial/dongle-combination other people describe as working.  Therefore i had to look for an alternative:

I spent my weekend on figuring out how to communicate with my jk-b2a24s20p via python-bleak via Bluetooth. Good news: I think i've already manages the hardest part and have all relevant readings (threshold & protection-voltages, per-cell-voltages and so on in a python-dict. 

In the next days i would have a closer look at your dbus-serialbattery implementation and how to fit this in. I would assume the easiest way would be to create another Battery-Subclass as a wrapper for the functions i wrote today in etc/dbus-serialbattery/jkbms.py e.g. jkbms_ble.py, so that while running dbus-serialbattery simply wouldn't only poll all the currently implemented serial-connections, but would also try to build up a bluetooth-connection and poll data from there.

Is this a/the way to go? Anything else to consider?

Louisvdw:
[...]

PR #79, opened 2022-01-09
Renogy

Louisvdw:


Louisvdw:
Hi @gambleben 
If possible can you test https://github.com/Louisvdw/dbus-serialbattery/releases/tag/v0.12b4 to see if that still works with your version of the Renogy.

gambleben:
Hi Louis,

I'll take a look and try to test soon.  I finally got to install my
batteries in a new travel trailer with the Cerbo a couple months ago.  I
have 3 batteries so I created a fork and made a handful of changes to fix
bad readings, work correctly with Python 3, and handle my multiple
batteries (by making them look like a single battery with 12 cells).  I
also had to lengthen the polling interval as 1 second wasn't quite enough
time to refresh all 3 batteries for all of the values, at least in the way
that I'm pulling them right now.  I suspect the reason the addresses for
the batteries are different is that I also use the Renogy bluetooth monitor
and I've found that it goes through a process to discover any attached
batteries and assign them new modbus addresses when it is first plugged in.

I also found full, correct documentation for all of the addresses, so I
planned on adding in the relevant alarms when I get time.

I'd like to get your input on the best way to handle multiple Renogy
batteries.  I think we would want to have a routine that scans the
potential address range when the driver is initialized, adds a new battery
for each address that responds, and then the driver would loop through and
poll them as normal.  What I'm not sure of is how to add the new batteries
individually.

gambleben:
Are you including minimalmodbus with the project now?  That would honestly make implementation of the full Renogy spec much easier.

Louisvdw:
After v0.11 the driver needed some changes to be able to work with the structure changes in VenusOS 2.80+ and also for python3. These are not backward compatable anymore which ment we can assume everything will be in Python3 and thus I could include Minimalmodbus. 
So you can assume it will be available, and use it.

Louisvdw:
I have been playing with the multiple battery (VenusOS cal them Multi Banks) feature a bit. 

Here is how I see the multi battery working. 
The serialbattery will create it's own custom path in the dbus.
Each battery/BMS will read it's data and "store" that under this custom dbus path.
There has to be a master service that will be doing the publish to the dbus. It will combine all the data from the custom dbus path for each bms and then publish this to the victron paths. 

There are also a few new batteries that I am working on that also need similar changes and it could also help with easier setup up for users. What happens with the SBMS is that the battery is not polled for data, but it just publish it so you need to be listening. And so it needs some master service as well.

The master service could also be what is installed by default. And then the user can do a battery detection so that the driver does not need to run though all the possible battery types each time. They can then also setup that battery with the settings needed which will move all the settings out from utils.py, but this needs a GUI 

There is also a ticket open to[ add bluetooth support](https://github.com/Louisvdw/dbus-serialbattery/issues/13) for the driver. One of the reasons I have started with that is the set up that is needed to be detected (addresses) and I think the driver needs a UI that can handle that. Seeing as you have done a bit of the bluetooth already maybe we can start with that using any hard codes for now and then move to a GUI and other BMS step by step?

Issue #68, opened 2021-11-05
Show all Cells within the battery on console detail screen


ddarek2000:
My current BMS can either be connected to BT or Serial connection to VenusOS.  I get to see all cell voltages in BMS , but would like to have the BMS sending the low and high cell to the VenusOS to control charge / discharge of the battery.

I would like to display all 16 cells in the detail, and have this accessible using the Console.

I have decided to return my current BMS and replace with one that supports dual connection.  This functionality would still be very useful for single connection devices.

Louisvdw:
What benefit do you see in having all the cell voltages available vs. just the min/max cell voltages that VenusOS currently support?
I know it will be nice to see. I am just trying to guage priority and if this is only a nice to have or has some benifit not currently available elsewhere.

ddarek2000:
The detail about batteries don't have concrete benefits.  I find it is easier to see if the min/max when compared to the average.  It is easier for me to interpret the state of each battery cell that way.  Even with lower priority , I might take this one up on myself to complete.

Louisvdw:
The [Chargery BMS driver ](https://github.com/Tobi177/venus-chargerybms) has support to display all the cells. The QML will need to be changed to add to the display (see gui\qml). 
In Victron land anything that is None will not be displayed, and this would be preferred as the driver support batteries with few to many cells (3-32 mostly). 
Also if we bring in the extra battery banks then this will complicate the display more. Something to keep in mind.

neoneddy:
+1 .. one of the things I look a the most in the BMS app right now (about to actually connect this to Venus soon) is individual cell voltages.   it let's you know if a particular cell is going rogue, or is it just .05v deviation from the rest.

With the QML, are you saying if someone wanted that data, just add the pointers / references in the QML and it will work?

Louisvdw:
@neoneddy , The VenusOS only contain the min/max cell voltages. If you have a rogue cell you can see it as either the min or the max will have larger differenaces from the other. This is also there the BMS min/max cell graph come in.

However, if we would want to show all the cell voltages, then these will need to be populated as non standard extra fields on VenusOS, and we also need to create/edit a view to display them with the QML.

So no, it will not work only adding QML.

![image](https://user-images.githubusercontent.com/26077665/142806257-2bfa5c4e-7761-44dd-abcc-e4c172bcc1fa.png)

neoneddy:
@Louisvdw Ok, makes sense. I think the min/max will work fine.

[...]

PR #87, opened 2022-02-03
Cell Voltages

piapiacz:
Response to issue : Show all Cells within the battery on console detail screen #68
Added publishing of cell voltages and cell balancing to dbus.
Added qml files for visualisation of cell voltages
qml files are installed by running "sh /data/etc/dbus-serialbattery/installqml.sh"
WARNING! - by running the "installqml.sh" is the existing file "/opt/victronenergy/gui/qml/PageBattery.qml" overwritten
Venus - Cell Voltages
The etc/dbus-serialbattery/qml/PageBatteryCellVoltages.qml file is for 16 batteries

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -155,6 +155,20 @@ class Battery(object):
         cell_no = self.get_max_cell()
         return cell_no if cell_no is None else 'C' + str(cell_no + 1)
 
+#cell voltages - begining
+    def get_cell_voltage(self, idx):
+        if idx>=min(len(self.cells), self.cell_count):
+          return None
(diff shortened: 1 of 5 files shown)


Louisvdw (approved):

Louisvdw:
The feature has not been dropped. I just missed the new files (installqml and QML folder) in the build script.

Louisvdw:
For the Daly battery the cell values does not work for all models it seems and then the min/max cell info is wrong. So for the Daly BMS the cell details are disabled until we can sort that out.

SamuelBrucksch:
Daly single voltage cells are fixed here:
https://github.com/Louisvdw/dbus-serialbattery/pull/162

PR #90, opened 2022-02-06
Add support for cell voltages and TimeToSoC items on dbus

CaptKrisp:
-battery.py
    Help prevent terminating when BMS doesn't return valid SoC data.
- utils.py
    Add config variable: BATTERY_CAPACITY
-dbushelper.py
    Add support for individual cell voltages.
    Add support for TimeToSoC for 5% SoC.
        All times will be positive regardless of whether charge/discharge.
        When charging and SoC is >= to a SoC percentage that percentage value will be None (or "-").
         When discharging and SoC is <= to a SoC percentage that percentage value will be None (or "-").
        If not charging or discharging all TimeToSoC will be None (or "-").
-daly.py
    Improve reliability of data read from Daly BMS
    Add cellvolt_buffer to read cell voltages and to accommodate up to 48 cells (Daly current max supported cells).
    read_soc_data attempts 3 times to read soc_data from Daly before failing.  This greatly increases the read reliability of some Daly devices.
    Add function read_alarm_data for most previously defined dbus Alarms items.
    Add function read_cells_volts to read read all individual cell voltages for up to 48 cells. Tested only with 4 cell BMS.
    Add logger.warning messages to understand which dataset read failed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 venv
 venus-data.tar.gz
 BMS-trials
-etc/.DS_Store
-etc/dbus-serialbattery/.DS_Store
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 venv
 venus-data.tar.gz
 BMS-trials
+etc/.DS_Store
+etc/dbus-serialbattery/.DS_Store
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -99,6 +99,10 @@ class Battery(object):
         # Start with the current values
 
         # Change depending on the SOC values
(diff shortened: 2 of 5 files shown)


Louisvdw (approved):
The value BATTERY_CAPACITY = 50 does not seem to be used.
Cell voltage is already included in a earlier Pull Request which I will merge with this.
Thanks for the contribution

CaptKrisp:
> The value BATTERY_CAPACITY = 50 does not seem to be used.

Thanks for catching that.  It is utilized in the get_settings, which I forgot to include the the repo copy.  It is there now. Shall I issue another PR for the change?

Louisvdw:
yes just create a new PR with the missing code and I will add it in

PR #92, opened 2022-02-07
Correct variable name and improve serial read stability

CaptKrisp:
-daly.py
Include voltage validation checks in read_soc_data
-battery.py
Replaced min_voltage with max_voltage in get_max_cell_voltage.
-utils.py
Increase serial read stability by allowing 3 attempts to open serial channel.


Louisvdw (changes requested):
I am happy with the fixes in battery.py and daly.py, but don't want to change the read serial data in utils.py with the retries.
I've had that before it breaks other things and cause problems between the different BMS implementations.

The reason that the Daly BMS is giving unstable results and not always returning the results is because of how daly implemented the protocol and how long it takes to send and receive all the bits. It has a lot of overhead for each command and this all runs on 9600 baudrate.

You will notice that the Daly driver sets the poll_interval = 2000 (2 seconds) and this was enough time to read all the stats without the alerts (one of the reasons I left the alerts out initially as there were not enough time to get them all).
Instead of the retries just increase this to 4000 and it should be enough time the get all the responses back in time. Give it a test and if you are happy remove the untils.py change from your PR. 

Something I was thinking about was to on every loop update the SOC, Voltage and current, but then alternate between the cell voltages and the alerts. Then you should be able to keep the poll_interval at 2000. That would be a bigger change to implement.

CaptKrisp:
What you said about the timing makes sense.
I have very little experience in python programming during the past 4 decades and still learning the nuances.

Also have very little experience using the PR flow in GitHub so hopefully I am following the proper flow.

CaptKrisp:
I submitted the requested changes and a couple other enhancements to the TimeToSoC routines.
I see that you have re-organized the code so it's causing some conflict with my changes.  I'm not sure how to pull your changes into my local so that I can resolve the conflicts.  Can you guide me?  Thanks!

Louisvdw:
Hi @CaptKrisp 
The easiest way to get all the updates is to use the github page of [your fork ](https://github.com/CaptKrisp/dbus-serialbattery) and then instead of the contribute button select the Fetch upstream from the drop-down.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

If you use the contribute again it should update your PR automatically.

You will see in the latest changes that I made some cleanup on the TimeToSoC  and moved it to battery.py as that is where most battery calcs are done.

CaptKrisp:
Hi @Louisvdw
Thanks very much for your time and guidance.
I have been using the GitHub Desktop app mostly and it doesn't have that option.  So for now I manually merged your code changes with my local repo to resolve the conflicts.

While reviewing your modifications I learned about the existence of the "deltatime" function, that exciting for me.
But also I noticed your moving primary code to the battery.py file introduces an inefficiency by causing the calculation of crntPrctPerSec to be performed for every SoC step (that would make for 32 extra calculations for the same value with all 11 SoC steps).  One way around that is to calculate the value in dbushelper then pass the value to the get_timetosoc function.  With this PR the code remains at the bottom of the publish_dbus function so that the code in get_timetosoc function can remain as you created it until a final decision on how you want to structure it is made.

But after adding that small change I realized that there are some other enhancements (primarily for user configuration/control) that I would very much like to have included and placing those in battery.py would also reduce efficiency a bit.

I'm not sure if this is the best place to discuss my code changes for consideration.  If not I would be pleased if you directed me to a better place.  For now, here is a summary of the changes I would like you to consider accepting.  These are all based on the following user config variables (with descriptions, examples and default values):

# Set of SoC percentages to report on dbus. The more you specify the more it will impact system performance.
#TIME_TO_SOC_POINTS = [100, 95, 90, 85, 80, 75, 70, 65, 60, 55, 50, 45, 40, 35, 30, 25, 20, 15, 10, 5, 0]		# Every 5% SoC
#TIME_TO_SOC_POINTS = []	                # No data set
TIME_TO_SOC_POINTS = [100, 90, 75, 50, 25, 20, 10, 0]

# Specify TimeToSoc value type: 1 = as seconds; 2 = as string (days, HR:MN:SC)
TIME_TO_SOC_VALUE_TYPE = 1

# Specify how many loop cycles between each TimeToSoc updates
TIME_TO_SOC_LOOP_CYCLES = 5

# Include TimeToSoC points when moving away from the SoC point.  These will be as negative time. Disabling this improves performance slightly.
TIME_TO_SOC_INC_FROM = False
#TIME_TO_SOC_INC_FROM = True

# Select the format of cell data presented on dbus.
BATTERY_CELL_DATA_FORMAT = 1                   # Format: /Voltages/Cell# (also available for display on Remote Console)
#BATTERY_CELL_DATA_FORMAT = 2                   # Format: /Cell/#/Volts
#BATTERY_CELL_DATA_FORMAT = 3                   # Both formats 1 and 2

Thanks so very much for all the time and effort you put into this project!

Louisvdw:
The PR is a good place to discuss as we have already been talking here. It just have other things happening in between the chats as well. The better plcae might be to create a new topic in the Discussions tab.

If you use Github Desktop you can use the View in Github option like in theis picture to go to your fork on the website.
image

There on the right you will find the Fetch and Merge option which will bring any changes from the original repo into your version. 
image

My suggestion would be to split your PR into 2. One for the fixes to Daly and other pieces and another for the new TimeToSoc changes. I will create a branch to TimeToSoc which you can work on and then this will not hold up any other changes and updates.

A good place for all the settings for TimeToSoc would be in the utils.py as this is the place that most would change settings in for the BMS.
I like the options you have added. 

For the calculations I was thinking it might actually be better to calculate the capacity per second when the battery starts up. The capacity will not change while it is running, so you can do this in get_settings() function

CaptKrisp:
Thanks for helping me get up to speed with GitHub.
The Fetch upstream button gives me a different result.
Screen Shot 2022-02-09 at 12 47 57 PM

I assume the way to split a PR is to first close.  The results of a quick search say to choose close from the Pull request list but my list is empty.  There is a "Close with comment" button at the bottom of this PR.  Is that the correct action for me or am I confused with the GitHub workflow?

Louisvdw:
Hi @CaptKrisp 
The Close Pull Request button will change to Close with comment as soon as you type anything in the comment box. Anyone is fine to use and a comment is always good so that everyone knows why things happen. 

When you do the Fetch Upstream you will get the latest of the upsteam repo, and as there are changes to the same areas in the code it has the conflict (you manually applied to the changes, but git now does not know which is the correct one). The pull request will be on your repo for the upstream changes, and you can then make sure it all works and accept and apply the PR to your own repo.

Louisvdw:
When you fetch upstream now you will see I created a TimeToSoc branch. Make all those changes in that banch and do the PR for those changes also to that branch. I have disabled the call in the master for the time being.

CaptKrisp:
Closing to split PR

Issue #165, opened 2022-07-30
Sometimes cell voltage is wrong on Daly


SamuelBrucksch:
Follow up to #160 

Sometimes wrong values come in. Here is such a sample:
```
@4000000062e4d4e2223bb9ac INFO:SerialBattery:cell 13 has invalid value: -23.295
@4000000062e4d4e2224cd0ac INFO:SerialBattery:Frame: 0xa5, 0x1, 0x95, 0x8, 0x1, 0xd, 0x29, 0xd, 0x33, 0xd, 0x32, 0x90, 0x89, 0xa5, 0x1, 0x95, 0x8, 0x2, 0xd, 0x2f, 0xd, 0x2e, 0xd, 0x33, 0x90, 0x8c, 0xa5, 0x1, 0x95, 0x8, 0x3, 0xd, 0x33, 0xd, 0x33, 0xd, 0x33, 0x90, 0x96, 0xa5, 0x1, 0x95, 0x8, 0x4, 0xd, 0x33, 0xd, 0x32, 0xd, 0x2f, 0x90, 0x92, 0xa5, 0x1, 0x95, 0x8, 0x5, 0xd, 0x30, 0xd, 0x34, 0xa5, 0x1, 0x95, 0x8, 0x6, 0xd, 0x32, 0xd, 0x34, 0xd, 0x31, 0x90, 0x97, 0xa5, 0x1, 0x90, 0x8, 0x2, 0x1c, 0x0, 0x0, 0x76, 0x56, 0x2, 0x86, 0xb0
```

Will check this. A first fix is to properly check for the low voltage value here:
https://github.com/Louisvdw/dbus-serialbattery/pull/166

Will investigate later if i can improve the byte parsing, maybe there was an error in the order that can be detected earlier.

SamuelBrucksch:
Data that came in:
```
0xa5, 0x1, 0x95, 0x8, 0x1, 0xd, 0x29, 0xd, 0x33, 0xd, 0x32, 0x90, 0x89
0xa5, 0x1, 0x95, 0x8, 0x2, 0xd, 0x2f, 0xd, 0x2e, 0xd, 0x33, 0x90, 0x8c
0xa5, 0x1, 0x95, 0x8, 0x3, 0xd, 0x33, 0xd, 0x33, 0xd, 0x33, 0x90, 0x96
[...]

PR #166, opened 2022-07-30
Properly get cell number based on message frame for Daly and filter invalid values

SamuelBrucksch:
the check did not filter out incorrect values before.

related to https://github.com/Louisvdw/dbus-serialbattery/issues/165

### DIFF
--- a/etc/dbus-serialbattery/daly.py
+++ b/etc/dbus-serialbattery/daly.py
@@ -229,7 +229,7 @@ class Daly(Battery):
             buffer[2] = self.command_cell_volts[0]
 
             maxFrame = math.ceil(self.cell_count / 3)
-            lenFixed = (maxFrame * 13) # 0xA5, 0x01, 0x95, 0x08 + 1 byte frame + 8 byte data
+            lenFixed = (maxFrame * 12) # 0xA5, 0x01, 0x95, 0x08 + 1 byte frame + 6 byte data + 1byte reserved
 
             cells_volts_data = read_serialport_data(ser, buffer, self.LENGTH_POS, self.LENGTH_CHECK, lenFixed)
             if cells_volts_data is False:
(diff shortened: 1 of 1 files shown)


etc/dbus-serialbattery/daly.py
```diff
@@ -227,8 +228,8 @@ def read_cells_volts(self, ser):
             buffer[1] = self.command_address[0]   # Always serial 40 or 80
             buffer[2] = self.command_cell_volts[0]
 
-            maxFrame = (int(self.cell_count / 3) + 1)
-            lenFixed = (maxFrame * 12)
+            maxFrame = math.ceil(self.cell_count / 3)
```
SamuelBrucksch, line 231:
```suggestion
            (int(self.cell_count / 3) + 1)
```

etc/dbus-serialbattery/daly.py
```diff
@@ -3,6 +3,7 @@
 from battery import Protection, Battery, Cell
 from utils import *
 from struct import *
+import math
```
SamuelBrucksch, line 6:
```suggestion
```

etc/dbus-serialbattery/daly.py
```diff
@@ -227,8 +227,8 @@ def read_cells_volts(self, ser):
             buffer[1] = self.command_address[0]   # Always serial 40 or 80
             buffer[2] = self.command_cell_volts[0]
 
-            maxFrame = (int(self.cell_count / 3) + 1)
-            lenFixed = (maxFrame * 12)
+            (int(self.cell_count / 3) + 1)
```
SamuelBrucksch, line 230:
```suggestion
            maxFrame = (int(self.cell_count / 3) + 1)
```

SamuelBrucksch:
@Louisvdw this improves cell voltage reading for Daly, to calculate the cell number based on the position in the frame.

However as it is still too unreliable, i disabled cell voltages detection for now.

PR #168, opened 2022-08-01
Fix cell numbers for daly

SamuelBrucksch:
Looks like i calculated cell number wrongly, with my CAN implementation i currently work on this was obvious, no idea why i did not see it on the serial version.


towatai:
Dont work with my 8S Daly over UART. in this new version, lines 67 and 68 are commented out. I had to re-enable them to get a display at all.

SamuelBrucksch:
Yes it is disabled as the serial connection seems to be quite unreliable. So it actually works but it is turned off. I had better results when only reading min/max values. This was also the reason for me to implement can in #169

PR #182, opened 2022-08-16
Charge voltage control

pawcio50501:
Adds possibility to dynamically control charge voltage.

FLOAT_CELL_VOLTAGE = 3.35 - voltage kept after MAX_VOLTAGE_TIME_SEC 
MAX_VOLTAGE_TIME_SEC = 15*60 - how long preserve max charge voltage
SOC_LEVEL_TO_RESET_VOLTAGE_LIMIT = 90 - when soc drops below this level, max charge voltage will be allowed again


pau1phi11ips (changes requested):
Ignore that, I see you reversed the commenting out.

pawcio50501:
> Edit: Ignore that, I see you reversed the commenting out.

Yes, I have jkbms only on my main brach.
The one used for the PR looks slightly different.

Louisvdw (approved):

Issue #181, opened 2022-08-15
Calculate remaining capacity if the BMS does not support that field


Louisvdw:
Many BMS/batteries does not publish the remaining capacity (capacity_remain). 
If the value does not exist, calculate it in Battery using the battery capacity and the SOC %.

Louisvdw:
It's in the Daly protocol but dont return valid values. So we can't use that command.

PR #204, opened 2022-09-09
ESC GreenMeter

Louisvdw:
GreenMeter implementation.
The Greenmeter must have a MODBUS address of 1


Louisvdw:
Remaining Capacity is calculated from the SOC if it cannot be read from the BMS
Fix #181

Issue #197, opened 2022-09-07
Renogy Smart Lithium iron phosphate with self-heating function cannot be detected


janderholm:
**Describe the bug**
Renogy Smart Lithium iron phosphate with self-heating function cannot be detected.

**To Reproduce**
Connect the battery and restart the victron gateway. It is never found.

**Expected behavior**
It is stated that Renogy batteries work.

**VenusOS (please complete the following information):**
 - Device type: Raspberry Pi 3
 - Firmware Version: 2.80

**Battery/BMS (please complete the following information):**
 - BMS/Battery type: Renogy Smart Lithium Iron Phosphate 12v 100Ah battery w/ self-heating function.
 - Cells: 4
 - Interface: USB-RS485

**Additional context**
This battery can be connected to successfully using the following module:
https://github.com/rosswarren/renogymodbus
I don't know if this battery is unsupported or if there're bugs in the integration, but I'll try to look into the issue. Perhaps this battery with heating is responds slightly differently.

janderholm:
I've managed to get it connected by changing address from 0 -> 247 and disabling the firmware version in the current Renogy integration.

But there are a few readings that are incorrect. Amp hours seem to be multiplied by 1000. How correct is the original implementation?

For example the firmware parsing code is broken because it tries to make a float out of "b'01'.b'31'" caused by an incorrect string conversion, so I don't think anyone have tried to use the integration on a current Venus OS (since python3 at least).

I'd like fix things but can only do that so they work on my batteries. So I'd need someone to test it on the original batteries as well.

Louisvdw:
Hi @janderholm 
I created a branch for you https://github.com/Louisvdw/dbus-serialbattery/tree/renogy-fixes
You can check it out and make your changes. When you are done create a Pull Request and I can make a beta build from that that other Renogy users can test and give us feedback.

Louisvdw:
It would be great if anyone can give the new beta a test that include changes for this issue ad give us feedback.
https://github.com/Louisvdw/dbus-serialbattery/releases/tag/v0.12b4

gambleben:
I recently forked the code and made some changes to work with Python 3 and the latest release of Venus, but haven't merged those changes back in.  Looks like you beat me to it, but I'll review my code and do some testing to see if there's anything else that should be fixed right away.

I also want to share the documentation for the Renogy modbus that I found: 
[...]

PR #205, opened 2022-09-10
Add support for Renogy RBT100LFP12SH

janderholm:
Implements #197 

May have fixed incorrectly reporting of total and current amp hours for other Renogy batteries as well.
I need some help to test and see that this works with currently existing batteries also. Specifically check that "Installed / Available capacity" in the "Details" menu is correct.

 It should report something like 100Ah |  89.5Ah. This was previously off by 1000 and reported as 100 000Ah | 89 500 Ah.

Also make sure that "Hardware Version" and "Firmware Version" under "Device" report something useful.


janderholm:
It's a bit tricky to test right now because master isn't working properly. minimalmodbus.py is missing from buildfiles.lst. But even when adding that, I have to manually start it using:

python /opt/victronenergy/dbus-serialbattery/dbus-serialbattery.py /dev/ttyUSB1

so something seems wrong with the automatic start.

Louisvdw:
I'll make a fix for the mising file in the build. Do you see anything happening in the serialbattery log files for USB1 ?

janderholm:
Nope, nothing. I have noticed that the execute bit is removed from all files in create_release.sh which I suspect is a problem for eg. start-serialbattery.sh and all other .sh files.

It works when I start off using the latest release and simply coping renogy.py and adding the addresses in dbus-serialbattery.py so I'm pretty sure it must be something added between that and now.

Louisvdw:
There are a few things to remember for testing.

When you copy files from local PC to the GX all attributes are removed. The builds that is released has all the attributes corrected and also fixes any line endings that could happen from Windows to the GX. The best option is to install a released build, and then only copy the files that you change over. This is most likely your issue.

The driver is installed at /data/etc/dbus-serialbattery/ and then copied over to /opt/victronenergy/dbus-serialbattery when ever the GX reboots. The /data/ folder is the only one where files and data stay with victron updates, which is why it is done this way. So if you edit /opt/victronenergy/dbus-serialbattery directly then it will be overriden at reboot.

So how I do it is to always start with a installed version. Then I copy/make my changes at /data/etc/dbus-serialbattery/ and first test it with a manual call. I disable serial-starter on that port first else you . 
Then when I am ready I will run 
> sh /data/etc/dbus-serialbattery/reinstalllocal.sh

which will move any changes to the /opt/victronenergy/dbus-serialbattery folder.

janderholm:
Right, copying the files and running it manually does work. But using "create_release.sh" to generate a tarball and installing it as per the instructions (either using installlocal.sh or copying it to a flashdrive) does not currently work.

Louisvdw:
Forgot to mention that you need to run the create_release.sh on a Linux machine. Those commands are not available on Windows. But you only need to to that if you want to make a build for release.
You can just use the latest b3 build (it's been updated to include the minimalmodbus module and then update with your changes you did in the renogy.py from your PC.

janderholm:
I do use Linux, but for whatever reason the tarball that I create has other permissions than what you make.

But I did install b3 using a flashdrive, copied over renogy.py and dbus-serialbattery.py rebooted and now it works well :)

So I believe anyone else with Renogy batteries can try it out!

janderholm:
I have found the documentation by Renogy for their batteries and apparently the default address is 247 (F7). They also respond to 255 (FF) as a broadcast address.

It is possible to set the address of a battery to something else so that they can be placed on the same bus and I think the renogy controller does this. So I believe the address previously used (0x30) may have been assigned by a Renogy controller.

Louisvdw (approved):

gambleben:
@janderholm - do you use the Renogy bluetooth monitor?  I have it, and using serial port monitoring discovered that it goes through a process of discovering all attached batteries and then assigns them new modbus addresses when it is first plugged in.  247 is the default address, and the bluetooth monitor starts assigning new numbers at hex 30 (ASCII 0 or decimal 48).
Ideally, I think we'd have a simple routine to poll the valid range of addresses and discover all the connected batteries, or would use the broadcast address to discover attached batteries.

janderholm:
I don't have the bluetooth monitor. How does it manage to set the address of the batteries without conflicts?

I have the PDUs for the batteries they released and can set and read the address to whatever, but since by default they respond on the same address any messages get garbled. Ideally, the batteries could be made to respond with their ID by broadcasting on address FF but I have issues with them talking over each other. 

I think an alternative approach to polling would be to simply add a toml config file which lists all batteries used and their parameters, effectively replacing the loop that is currently there.

PR #219, opened 2022-09-28
VenusOS 2.91 RW Mount on /

damienheiser:
Update reinstalllocal.sh with Read/Write remount for /

Now that / is Read Only, scripts were not running, this is a fix for that.

### DIFF
--- a/etc/dbus-serialbattery/reinstalllocal.sh
+++ b/etc/dbus-serialbattery/reinstalllocal.sh
@@ -3,6 +3,10 @@
 DRIVER=/opt/victronenergy/dbus-serialbattery
 RUN=/opt/victronenergy/service-templates/dbus-serialbattery
 OLD=/opt/victronenergy/service/dbus-serialbattery
+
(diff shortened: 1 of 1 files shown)


etc/dbus-serialbattery/reinstalllocal.sh
```diff
@@ -3,6 +3,10 @@
 DRIVER=/opt/victronenergy/dbus-serialbattery
 RUN=/opt/victronenergy/service-templates/dbus-serialbattery
 OLD=/opt/victronenergy/service/dbus-serialbattery
+
+#/dev/mmcblk1p3 mountpoint / was made read only in 2.91 release
+mount -o remount,rw /dev/mmcblk1p3 /
```
Louisvdw, line 8: Thanks for sharing your changes. Any reason you did not use the standard Victron remount script?
> /opt/victronenergy/swupdate-scripts/remount-rw.sh

Louisvdw (approved):

PR #238, opened 2022-10-05
 Changes to be committed: implemented constants for charge and dischare limits

corneel27:
modified:   etc/dbus-serialbattery/battery.py
	modified:   etc/dbus-serialbattery/utils.py

in utils.py added constants for charge and discharge limits 
in battery.py implemented these constants


Louisvdw (approved):
Thanks

PR #252, opened 2022-10-11
Settings for UI

Louisvdw:


etc/dbus-serialbattery/dbushelper.py
```diff
@@ -67,11 +66,11 @@ def setup_vedbus(self):
         # Create the mandatory objects
         self._dbusservice.add_path('/DeviceInstance', self.instance)
         self._dbusservice.add_path('/ProductId', 0x0)
-        self._dbusservice.add_path('/ProductName', 'SerialBattery(' + self.battery.type + ') v' +
-                                   str(DRIVER_VERSION) + DRIVER_SUBVERSION)
```
WaldemarFech, line 71: I would keep this. It's good to see the installed driver-Version.

WaldemarFech (commented):
not finished yet my review - just got no time xD

WaldemarFech (approved):
check :)

etc/dbus-serialbattery/dbushelper.py
Louisvdw: @WaldemarFech 
The problem is that the version clutters up the screen a bit (it gets too long). I played with the idea to replace the battery.version (firmware version if the BMS support that) with the driver version completely and not just when it is not populated. There is also a ProcessVersion path that we could potentially populate but I have not traced where that value will be displayed. 
In the end we will still display the driver version in the GUI somewhere. (it is also in the logs but needs to be in the GUI also).

Issue #239, opened 2022-10-06
Renogy randomly disconnects (v0.13)


TimGFoley:
While watching the Device List in the remote console, the Serialbattery will randomly show disconnected, it will reconnect after a few moments (less than a minute, will time next it happens).

Will continue to monitor to see if I can find a cause or pattern.

---

Venus GX (Victron OEM)
Firmware Version v2.91

Battery:
Renogy Smart Lithium 100AH
Cells: 4
Interface: Renogy USB to RS485 OEM Cable (no other devices connected to battery)

Copy of current log when it happens here:

@40000000633e00a2209ddc74   File "/opt/victronenergy/dbus-serialbattery/renogy.py", line 148, in read_temp_data
@40000000633e00a22147a164     self.temp2 = unpack('>H',temp2)[0] / 10
@40000000633e00a22147c48c TypeError: a bytes-like object is required, not 'bool'
@40000000633e00d0198ecefc INFO:SerialBattery:Starting dbus-serialbattery
@40000000633e00d019acea04 INFO:SerialBattery:dbus-serialbattery v0.13
@40000000633e00d019ed90cc INFO:SerialBattery:Testing LltJbd
@40000000633e00d02b33518c ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d02b5c3a34 INFO:SerialBattery:Testing Ant
@40000000633e00d10416787c ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d1043c194c ERROR:SerialBattery:>>> ERROR: Incorrect Data
@40000000633e00d1044cede4 INFO:SerialBattery:Testing Daly
@40000000633e00d1153c8fc4 ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d1156950cc INFO:SerialBattery:Testing Daly
@40000000633e00d1276486ac ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d127a642cc INFO:SerialBattery:Testing Jkbms
@40000000633e00d215a6f9f4 ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d2168d0494 INFO:SerialBattery:Testing Sinowealth
@40000000633e00d22f7066cc ERROR:SerialBattery:>>> ERROR: No reply - returning
@40000000633e00d22fa211a4 INFO:SerialBattery:Testing Lifepower
[...]

Issue #102, opened 2022-03-09
Cannot read property 'hasOwnProperty' of null" when using node-red


xury77:
**Describe the bug**
On battery monitor using node-red 
When no alarm property of dbus message in null but should be 0 
"TypeError: Cannot read property 'hasOwnProperty' of null"
**To Reproduce**
Connect to node-red use Baterry monitor node and choose any type of alarm
**Expected behavior**
Payload output should be 0 

**Screenshots**

**VenusOS (please complete the following information):**
 - Device type: Cerbo, 
 - Firmware Version 2.82 large

**Battery/BMS (please complete the following information):**
 - BMS/Battery type: Daly
 - Cells: [e.g. 16]
 - Interface:  USB-RS232

mr-manuel:
@xury77 does the issue still persist? In the meanwhile there were a lot of changes in the driver and Victron Node.

xury77:
I stopped using Daly BMS in favor of my own with your dbus  battery mqtt controller. So I don't know is still persist.

PR #267, opened 2022-10-18
Constants for soc low

corneel27:
Added constants for soc low in utils.py
and implemented these constants in lltjbd.py constants


Louisvdw (approved):

corneel27:
If you have other items which I can contribute: I am glad to help!

Louisvdw:
Hi @corneel27 
Thanks for the help. 
There are 2 bugs that still need to be looked at #102 and #239 if you want to look at those.
Then we also need to give users the option of which of the BMS drivers we should test. Something that we can later use in the GUI  #269 and is similar to what you have already done.

corneel27:
Hello Louis,

The bugs #102 and #239 are difficult for me:

  * for #102 I have to install and check node-red. I have no experience
    with node-red
  * for #239 I need a Renogy bms and the usb-rs485 from renogy. Both are
    not possible for me in short time

But #269 seems better for me. So I wil give that a try

Cheers, Cees


Op 20-10-2022 om 14:37 schreef Louis Van Der Walt:

Issue #269, opened 2022-10-20
Move battery_types to utils.py


Louisvdw:
We also need to give users the option of which of the BMS drivers we should test. 
At least one or more BMS option(s) must be selected, or all of them.

For this the battery_types list in def get_battery_type (dbus-serialbattery.py) should be defined in utils.py
 battery_types = [
            LltJbd(port=_port, baud=9600),
            Ant(port=_port, baud=19200),
            Daly(port=_port, baud=9600, address=b"\x40"),
            Daly(port=_port, baud=9600, address=b"\x80"),
            Jkbms(port=_port, baud=115200),
            Sinowealth(port=_port, baud=9600),
            Lifepower(port=_port, baud=9600),
            Renogy(port=_port, baud=9600, address=b"\x30"),
            Renogy(port=_port, baud=9600, address=b"\xF7"),
            # Revov (port=_port, baud=9600),
            Ecs (port=_port, baud=19200),
            # MNB(port=_port, baud=9600),
        ]

PR #270, opened 2022-10-21
moved battery_types to utils.py

corneel27:
moved battery types to utils.py (as a list if dictionaries)
imported these types in dbus-serialbattery.py and changed the code in def get_battery_type
i have tested this update for Jbd: positive
I have NOT tested this update for other bms/batteries, special attention for bms with other baudrate and/or other address


corneel27:
hereby issue #269 could be resolved

Louisvdw (approved):

Issue #297, opened 2022-11-24
CCL/DCL depending on cell-voltage does not turn off charging completely, still overvoltage alarm


chaerle:
When using CCL/DCL from https://github.com/Louisvdw/dbus-serialbattery/pull/293 with mode "WaldemarFech" the code correctly sets the charge current limitation (CCL) to zero, if the highest cell voltage gets over 3.55V. The Victron Multiplus II 48/3000 however does not honor it precisely, and still keeps charging at 0.4A to 0.6A, thus still driving the battery into over-voltage-alarm.

The charge current in the Multiplus seems to be always higher than the CCL, which is not a big problem as long the CCL ist above zero. But if the charge limit is zero it should definitively have zero current to the battery to avoid over charging. 

I also observed that the LED in on the Multiplus keeps in "Bulk" mode and does not change into the "absorbtion" or "tickle". I don't know if this is correct or if switching the mode could solve the problem, because the Multiplus might use different charge current limitation algorithms in different modes. 

If I set the charging current in DVCC to zero it still have this 0.6A charge. But I was able to have zero charge current by lowering the maximum charging voltage. Interesting effect: After charging current dropped to zero, I increased the max charge voltage, but charge current kept zero.

The code should ensure that the charge current is really zero, either by switching off the charger, or changing the charging mode or altering the charge voltage. There might be also energy savings when the charger is completly off.

WaldemarFech:
hm.
so you see the CCL in the remote console set to zero in the serial battery - parameters menu?

first we have to find out weather it's a bug in our code or if it's a strange behavior of the internal Victron charge control.

i could change the code to print some debug messages in a log file, because I don't think, that i can reproduce it. here it's working fine.

if the code is not the problem, then we can think about workarounds. It's no problem to lower the max charge voltage with the CVL.

But first we have to figure out the reason for the continuous charge current.

Let me change a bit - see you here again.

chaerle:
> so you see the CCL in the remote console set to zero in the serial battery - parameters menu?>

Yes. Zero in CCL, but 0,6A still loading into the battery.

Louisvdw:
It won't work like you expect. There always will be a bit of current flowing into and out of the battery. I've done many tests on this question and talked to Victron people.

Firstly the chargers cannot accuratly limit/set the charge under 0.6A. Anything under that will be ignored except for the smallest inverters which might have a bit smaller value.
[...]

PR #312, opened 2022-11-30
Linear limits

Louisvdw:
This is a merge between the classic charge control management and the nice contributions from @WaldemarFech 

There are 3 separate limit options that can all be selected - Cell Voltage, SOC, Temperature. The lowest enabled limit will by applied
You can now choose how limits will be applied in Step or Linear mode.

Improved the calc relationship selection code with bisect()


rcrc1978:
I'm having a little problem on this one, the linear setting.  My CVL limit won't go above 13.4v with it set to True. Posted a little more info here [https://github.com/Louisvdw/dbus-serialbattery/issues/297#issuecomment-1381227950](url)
but it was marked closed.

Issue #360, opened 2022-12-19
installrelease.sh overwrites the existing /data/rc.local file


nc1066:
**Describe the bug**
Running installrelease.sh overwrites the existing /data/rc.local file. If rc.local is already reloading a custom driver then after the install of the dbus-serialbattery driver the custom driver no longer reloads on reboot. 

**To Reproduce**
run installrelease.sh with an existing /data/rc.local, original contents of rc.local are gone 

**Expected behavior**
Install of dbus-serialbattery driver should not overwrite /data/rc.local

**Screenshots**
NA

**VenusOS:**
 - Device type: Raspberry Pi 3 model B rev 1.2
 - Firmware Version 2.92

**Battery/BMS (please complete the following information):**
 - BMS/Battery type: EG4 Lifepower 
 - Cells: 8 (24V, 200Ah)
 - Interface: USB-RS485

**Additional context**
I'm loading a custom driver for an Epever CC. I use rc.local to recreate a symbolic link between /service/epever and /data/epever so the Epever CC driver will reload on reboot.

Issue #128, opened 2022-06-09
Installing this driver destroys a already installed one


stundenblume:
1.: Thank you for this very helpful driver!

**Describe the bug**
Installing this driver destroys a already installed one.

**To Reproduce**
1. working [dbus-mqtt-devices](https://github.com/freakent/dbus-mqtt-devices) driver
2. installing this driver
3. reboot
4. dbus-mqtt-devices driver doesn't work anymore

**Expected behavior**
Installing this driver should not influence other drivers.

**Solution**
Reinstalling dbus-mqtt-devices helped

**What could I find**
Your installation script replaced a already existing entry in `/data/rc.local`
Reentering it did not help. I had to reinstall dbus-mqtt-devices

Louisvdw:
This was a known issue for the autoinstaller that the VenusOS provides0. 
For serialbattery it will not be a major issue if the rc.local with it's commands don't exist anymore after installation, but then when you upgrade your VenusOS firmware next time you will have to install the driver again. 

The only way to get around this would be to not use the auto installer option, and then adjusting the ssh install script to not destroy any existing files.

stundenblume:
@ngardiner
the setup script of dbus-mqtt-devices [appends](https://github.com/freakent/dbus-mqtt-devices/blob/main/bin/setup.sh#L33) his line to `/data/rc.local`

For me everything is working new. But just to be aware that this can happen.

ngardiner:
> @ngardiner the setup script of dbus-mqtt-devices [appends](https://github.com/freakent/dbus-mqtt-devices/blob/main/bin/setup.sh#L33) his line to `/data/rc.local`
> 
> For me everything is working new. But just to be aware that this can happen.

That's going to work I'd nobody puts an exit 0 at the end of their script or runs any command which doesn't either fork or finish immediately, but we are at the point where people have to guess what other scripts are going to do.

kr0815:
i found this old topic, have exactly the same problems
In rc-local there was en entry that reinstalled dbus-serialbattery at each reboot - at the reinstall deleted my other entries there

I uncommented that entry - now i want to change max charge / discharge current in utils.py, but this has no effect?
utils.py says 20A/20A , Venus GUI says 5A / 60A , very strange
Do i really need a reinstall each time the unit reboots?

pos-ei-don:
Hm, sadly the installation deleted my rc.local without any backup. so i lost many settings.....

Louisvdw:
Fixed in #370

PR #370, opened 2022-12-28
Improve and fix install script, fix issue #360

chaerle:
Simplify install script
Fix wrong remove of $OLD 
Fix issue #360 
Force restart of running service without reboot


Louisvdw (approved):

Louisvdw:
also fix issue #128

PR #383, opened 2023-01-01
Bug while detecting non answering battery

baranator:
Stumbled about this while testing:

`error_count` is local to publish_battery() and there reset to 0 on every call. Therefore it will never reach the later values of 10 or 60 that indicate a problem. Possible solution: make error_count a class variable


Louisvdw (approved):
Thank you

PR #417, opened 2023-01-14
Update dbushelper.py

corneel27:
added LowSoc in dbus TimeToGo items


corneel27:
Made another with the correct dbus path

PR #439, opened 2023-01-28
Improved JBD BMS soc calculation

aaronreek:
The SOC reported by the JBD BMS is only integer precision. This results in steps of the charge/discharge current limit even when using the linear calculation method for charge current limit by SOC (I get steps of 5 amps per 1% SOC with my current settings). 

image

image

By calculating the SOC based on capacity_remain / capacity, we get a floating point number (this is presumably the same calculation the BMS is doing internally, just returning an integer). Using this method, the resulting charge/discharge limit has a smooth character.

image

image


aaronreek:
Just an update: I've been using the improved SOC calculation since I posted ~2 months ago, and it's working as expected.

Louisvdw (approved):

PR #440, opened 2023-01-29
Show MOSFET temperature from JKBMS

baphomett:
I had a JKBMS repeatedly cutoff discharge for an unknown reason. After some digging, finally found it was the MOSFET temperature limits that were set too low. But I had to get close to the BMS and open the JKBMS app to use Bluetooth to get the MOSFET temp because it was not showing up through the serial battery driver. 

Took the time to did through the byte stream of data that JKBMS publishes and found that it sends the MOS temperature as well. 

Had to make changes in the battery class to accept a new temperature attribute, also create a dbus path and add a GUI interface to the qml file. 

It shouldn't break anything, but I couldn't test using a BMS that is not JK.

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -76,6 +76,7 @@ class Battery(ABC):
         self.temp_sensors = None
         self.temp1 = None
         self.temp2 = None
+        self.temp_mos = None
         self.cells: List[Cell] = []
         self.control_charging = None
         self.control_voltage = None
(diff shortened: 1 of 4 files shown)


mr-manuel:
@Louisvdw could you please merge this? It's running fine since 3 weeks on my end. Do you need help with the open pull requests?

locke981 (commented):
Tested it succesfully in my environment including a JK BMS, the Mosfet temperature shows up in ve connect as expected.

mr-manuel (approved):
Is already running fine since months.

mr-manuel:
Since it's a small easy to understand PR I will merge it. The black lint errors will be fixed with my PR.

mr-manuel:
`str` variable assignment instead of `int` will be fixed with https://github.com/Louisvdw/dbus-serialbattery/pull/468/commits/10972b9249bcd75a6f79e9c802e93423e9ee27d1

PR #454, opened 2023-02-09
JKBMS Bluetooth - added battery details

mr-manuel:
- added: capacity
- added: MOS temperature
- added: charging switch status
- added: discharging switch status
- added: balancing switch status
- added: show if balancing is active and which cells are balancing
- added: cell imbalance alert

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -53,10 +53,12 @@ class Battery(object):
         self.soc = None
         self.charge_fet = None
         self.discharge_fet = None
+        self.balance_fet = None
         self.cell_count = None
         self.temp_sensors = None
         self.temp1 = None
         self.temp2 = None
+        self.temp_mos = None
         self.cells = []
         self.control_charging = None
         self.control_voltage = None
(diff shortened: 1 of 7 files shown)


baranator:
One suggestion: From the jkbms you can read `["settings"]["balance_trigger_voltage"]` maybe use this value (and a multiple of this value) rather than fixed values like 0.05 and 0.1 ?

mr-manuel:
That‘s a good idea. I‘m currently still gaining experience at which point the cell voltage difference is critical. What would you say when the cell difference should be warning and critical?

baranator:
> That‘s a good idea. I‘m currently still gaining experience at which point the cell voltage difference is critical. What would you say when the cell difference should be warning and critical?

Same here. No idea.

Critical-value however could be the one the bms has set as its trigger `["settings"]["balance_trigger_voltage"]` . And warning? maybe half that value?

mr-manuel:
Regarding the cell voltage difference I found now that 5% of the maximum cell voltage or 200 mV are considered as critical.
So I would use 200 mA or 5% of OVP as critical and 120 mA or 3% of OVP as warning. Over Voltage Protection is fetched from the BMS.

PR #466, opened 2023-02-16
jkbms_ble: 32s-Support and connection workaround for bt-fail

baranator:
- Support for newer jkbms that send frames with up to 32 cell infos.
- Also a workaround for the bug that causes the bluetooth-connection to fail after 15 hours and 22 mins


etc/dbus-serialbattery/jkbms_brn.py
```diff
@@ -11,7 +11,7 @@
 
 # zero means parse all incoming data (every second)
 CELL_INFO_REFRESH_S = 0
-DEVICE_INFO_REFRESH_S = 60 * 60 * 5  # every 5 Hours
+DEVICE_INFO_REFRESH_S = 60 * 60 * 10  # every 10 Hours
```
mr-manuel, line 14: Is there a reason why you chose 10 hours and not 12 hours?

baranator: tbh: That line can be removed completely as it not used anymore - device-info only contains non-changing data that do not need to be updated after connect.

That change comes from back then, when we tried to figure out where that crashes come from...

PR #513, opened 2023-04-06
Add dynamic voltage adaption

smurfix:
This PR adds a simple but very effective way to adapt a battery's voltage limits, based on the states of its individual cells.

To explain the algorithm, let's assume the configured maximum cell voltage is 3.5V. If the top cell is at 3.4V and all others are at 3.39 or so, the pack is well-balanced and we can (most likely) safely charge until they're all at or slightly below 3.5V. The maximum is thus `3.5V*n_cells`.

However, if these other cells are at 3.3V or so, we can only increase the total pack voltage by 0.1V before the top cell's voltage exceeds 3.5V. In this case the other cell voltages will not move much, if at all, due to the highly nonlinear charge curve of LiFePo4 cells. The maximum voltage the system can use is now `3.3V*(n_cells-1) + 3.5V`, or `3.3V*n_cells + 0.2V`, which is significantly less.

This code uses the voltage difference between the top cell and the configured maximum to determine how much to "discount" the *other* cells' contribution to the (additional) voltage which the battery can reach before the top cell goes over the limit.

This method is not perfect, but in combination with the existing current limitation it reliably prevents over-voltage cells even when the battery has never been balanced.

The same argument in reverse applies to the minimum battery voltage.

### DIFF
--- a/etc/dbus-serialbattery/dbushelper.py
+++ b/etc/dbus-serialbattery/dbushelper.py
@@ -323,11 +323,12 @@ class DbusHelper:
                 if self.error_count >= 60:
                     loop.quit()
 
-            # This is to manage CCL\DCL
+            # This is to manage CCL/DCL
             self.battery.manage_charge_current()
 
-            # This is to manage CVCL
+            # This is to manage CVCL and DVCL
             self.battery.manage_charge_voltage()
+            self.battery.manage_discharge_voltage()
 
             # publish all the data from the battery object to dbus
             self.publish_dbus()
@@ -395,8 +396,8 @@ class DbusHelper:
             "/Info/MaxDischargeCurrent"
         ] = self.battery.control_discharge_current
 
-        # Voltage control
-        self._dbusservice["/Info/MaxChargeVoltage"] = self.battery.control_voltage
+        self._dbusservice["/Info/BatteryLowVoltage"] = self.battery.control_charge_voltage
+        self._dbusservice["/Info/MaxChargeVoltage"] = self.battery.control_discharge_voltage
 
         # Updates from cells
         self._dbusservice["/System/MinVoltageCellId"] = self.battery.get_min_cell_desc()
--- a/etc/dbus-serialbattery/dbushelper.py
+++ b/etc/dbus-serialbattery/dbushelper.py
@@ -323,10 +323,10 @@ class DbusHelper:
                 if self.error_count >= 60:
                     loop.quit()
 
-            # This is to mannage CCL\DCL
+            # This is to manage CCL\DCL
             self.battery.manage_charge_current()
 
-            # This is to mannage CVCL
+            # This is to manage CVCL
             self.battery.manage_charge_voltage()
 
             # publish all the data from the battery object to dbus
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -78,7 +78,8 @@ class Battery(ABC):
         self.temp2 = None
         self.cells: List[Cell] = []
         self.control_charging = None
-        self.control_voltage = None
+        self.control_charge_voltage = None
+        self.control_discharge_voltage = None
         self.allow_max_voltage = True
         self.max_voltage_start_time = None
         self.control_current = None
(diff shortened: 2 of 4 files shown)


smurfix:
Converted to a draft as I'm doing this for a client and the whole thing is still entirely untested.

smurfix:
>  Should this PR replace the CVL for linear and/or step mode?

In my opinion the linear/step mode is still useful, and might even work better, for battery chemistries that are *not* LiFePo4.

However I do wonder why there's a step mode in the first place. Given the existence of internal battery/cell resistance, step functions that don't have a hysteresis will result in an oscillating, hence unstable, system.

But that's a different discussion.

mr-manuel:
Step mode is the oldest and it's only switching between max voltage and float.
Linear mode is applying penalties to the CVL, if one cell exceeds the cell voltage limit.
I think it would be much better to keep the "step mode" (default) and replace the linear with your calculations. I was thinking about the same. Then when all cells have the same voltage the CVL is switched to float voltage. https://github.com/Louisvdw/dbus-serialbattery/discussions/453

pos-ei-don:
I think this is veeery helpful!
Thank you very much.
Maybe this van be added as third option: "steps, linear, cellvoltagediff

smurfix:
I wouldn't decrease CVL in that case. CVL is the max safe voltage the battery can have; I don't think using it for anything else is good practice. If you want the system to stop charging, just turn off the `AllowToCharge` flag.

Taking enough current off the system so that the battery stays at the charge level you want it to be at is a problem best solved separately IMHO. Doing that is not the battery controller's job.

mr-manuel:
This was done since the MultiPlus cannot handle well small currents. The current is already reduced, but this helps the system to charge the battery at low current and let the balancer do it's job.
If you now add another similar possibility it complicate things for users. Already now great part don't understand the simplest settings so let us keep it as simple as possible.

pos-ei-don:
Its even me who does not understand the other settings in depth. 
But it seems that this New Option is simple enough for me. I am currently very happy.

Multiplus 10k in 3 phase is even worse with small currents.
Most of the time I prefer using "no phase  compensation", but  when the currents are very low I need to switch to the opposite.

smurfix:
OK, after fixing a bunch of stupid errors 🫢 this is now in productive use at my customer's site and (so far at least) seems to be working OK.

smurfix:
Anything I should do to get this merged?

mr-manuel:
Just wait :-) Louis is busy at the moment, but he gets back mid May. If you want I can merge it in the meanwhile with the `jkbms_ble` branch.

smurfix:
Thanks. Feel free to merge (and test the heck out of it of course).

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -395,8 +396,8 @@ def publish_dbus(self):
             "/Info/MaxDischargeCurrent"
         ] = self.battery.control_discharge_current
 
-        # Voltage control
-        self._dbusservice["/Info/MaxChargeVoltage"] = self.battery.control_voltage
+        self._dbusservice["/Info/BatteryLowVoltage"] = self.battery.control_discharge_voltage
```
mr-manuel, line 399: Is this working?

> `/Info/BatteryLowVoltage     <- Note that Low Voltage is ignored by the system `

See https://github.com/victronenergy/venus/wiki/dbus#battery

smurfix:
> If a cell reaches the max voltage limit, then the CVL should kept at the current voltage + a small amount, maybe 0.01 * cell count.

That's what my code does at the limit. I'm more concerned about what happens when you're below that.

I'm interested in "which charge/discharge voltage is safe", aka "how much more voltage than right now can the pack take without exceeding a single cell's top voltage". You can't get that value by looking at the cells in isolation. The "penalty" for a single cell that's at 3.4V strongly depends on whether all the other cells are at 3.3V or 3.4V or 3.45V.

I'm not at all interested in asking the BMS to reduce the max voltage once balancing is achieved because (a) IMHO that's not the job of the BMS, esp. when you have more than one battery, (b) which criteria are you going to use to re-raise the voltage?

mr-manuel:
(a) therefore we have this driver that manages the charging algorithm. If you haven't connected a BMS then the MultiPlus for example is managing the charging algorithm. As soon as you connect a BMS and select it under DVCC (Venus OS >v3.00), then the BMS is managing the charging algrithm. Before Venus OS v3.00 the first connected BMS is selected.

When I have multiple batteries, then a battery aggregator is needed, which act as the managing device for all batteries (only reliable possible with Venus OS >v3.00). This because Venus OS is using the data of only one device.

Since this driver is not acting as battery aggregator, but managing only one specific battery, it should mirror the needs that this battery has. Then a battery aggregator has to choose which CVL to set on the charging device. I would pick the smallest voltage.

grafik

(b) The voltage is recalculated every x seconds during max cell voltage reached. The voltage is raised, if cell voltage is below max voltage. Here you can see it very well, how it's currently working with the penalty voltages (even, if my charger cannot output more than 64 V, but indeed you see very good how it works):

grafik

Do you track your data do verify that it's working as expected?

mr-manuel:
Here another diagram of my change. I don't think your PR is still needed.

grafik

PR #582, opened 2023-04-26
fix dbus-daemon memory leak

seidler2547:
`poll_battery` is already called from the dbus mainloop in a separate thread. Opening a new thread here (and especially never closing it) will accumulate threads and leads to problems like `dbus-daemon` memory usage increasing over time, leading to eventual reboot

disclaimer: I have not tested this with a USB/serial connection, only Bluetooth

### DIFF
--- a/etc/dbus-serialbattery/dbus-serialbattery.py
+++ b/etc/dbus-serialbattery/dbus-serialbattery.py
@@ -57,11 +57,7 @@ logger.info("Starting dbus-serialbattery")
 
 def main():
     def poll_battery(loop):
-        # Run in separate thread. Pass in the mainloop so the thread can kill us if there is an exception.
-        poller = Thread(target=lambda: helper.publish_battery(loop))
-        # Thread will die with us if deamon
-        poller.daemon = True
-        poller.start()
+        helper.publish_battery(loop)
         return True
 
     def get_battery(_port) -> Union[Battery, None]:


mr-manuel:
I'm also testing it with a serial connection, but looks good!

seidler2547:
> > `poll_battery` is already called from the dbus mainloop in a separate thread
> 
> Where did you find that?

The `gobject.timeout_add` function sets a function (specified by callback) to be called at regular intervals.

Which means that `poll_battery` is not called by our application directly, it is being called at the specified interval from the mainloop. And I don't see why it should create a new thread at regular intervals, especially if it's not waiting for that thread to actually finish. If `poll_battery` takes longer to finish, it will always continue to create new threads.

I did test this for longer and it seems it's not a complete fix. I stayed stable for about one hour, then started to behave similarly to before. I still think it's an improvement though to remove the unnecessary thread creation.

mr-manuel:
Thanks

mr-manuel:
Now we can also remove this?

https://github.com/Louisvdw/dbus-serialbattery/blob/964ec17df69dbef95baf6d81f5153064aeae62c2/etc/dbus-serialbattery/dbus-serialbattery.py#L7

seidler2547:
Yes, we can also remove that. I have put a watchdog script in place now which checks the memory usage of `dbus-daemon` and restarts the dbus-serialbattery process as soon as it changes.

Here are the results:
swappy-20230429_195414
The zigzag curve on the left is how the memory usage usually developed - going down until a reboot (either forced by the system or scheduled by me externally).
Then on the evening of the 28th I add the watchdog script. You can see that the memory usage is almost completely stable now. It restarts about every 2h:

- 04:29
- 06:24
- 08:24
- 10:24
- 12:29
- 14:34
- 16:44
- 18:59

So all in all quite weird, but there seems to be a bug with either long-running BLE connections or long-running DBus connections.

PR #658, opened 2023-05-21
Add Supoprt for HeltecSmartBMS (YYBMS) using modbus via RS485 connection

### LINKED ISSUE #508: Add Support for Heltec BMS
**Is your feature request related to a problem? Please describe.**
I read on https://github.com/Louisvdw/dbus-serialbattery/wiki/BMS-types-supported that "JKBMS / Heltec BMS" are supported, so I tried to connect my Heltec Smart BMS 9-25S 100A RS485 but it is not detected by the dbus-serialbattery driver.

**Describe the solution you'd like**
According the docs I have (and what I spied on the RS485 with the scope) the BMS seems to talk Modbus RTU and is different from the JKBMS. So I propose to add support for this Heltec device. I don't know about other Heltec BMSs and if there is a variant that is compatible to JKBMS, but we might also want to update the docs accordingly.

ramack:
This would close #508 / #575

### DIFF
--- a/etc/dbus-serialbattery/bms/heltecmodbus.py
+++ b/etc/dbus-serialbattery/bms/heltecmodbus.py
@@ -1,0 +1,332 @@
+# -*- coding: utf-8 -*-
+# known limitations:
+#   - only BMS variants with 2 cell temperature sensors supported
+#   - some "interesting" datapoints are not read (e. g. registers 52: switch type, 62: bootloader and firmware version)
+#   - SOC not yet resettable from Venus (similary to Daly for support of writing SOC), but modbus write to 120 should be fairly possible)
+
+
+from battery import Protection, Battery, Cell
+from utils import *
+from struct import *
+import time
+import minimalmodbus
+from typing import Dict
+import threading
+
+RETRYCNT = 10 # the Heltec BMS is not always as responsive as it should, so let's try it up to (RETRYCNT - 1) times to talk to it
+    
+SLPTIME = 0.02 # the wait time after a communication - normally this should be as defined by modbus RTU and handled in minimalmodbus, but yeah, it seems we need it for the Heltec BMS
+
+mbdevs: Dict[int, minimalmodbus.Instrument] = {}
+locks: Dict[int, any] = {}
+
+class HeltecModbus(Battery):
+    def __init__(self, port, baud, address):
+        self.address = address
+        super(HeltecModbus, self).__init__(port, baud, address)
+        if address != 0:
+            type = "#" + str(address) + "_"
+        else:
+            type = ""
+        self.type = type + "Heltec_Smart"
+
+    def test_connection(self):
+        # call a function that will connect to the battery, send a command and retrieve the result.
+        # The result or call should be unique to this BMS. Battery name or version, etc.
+        # Return True if success, False for failure
+        logger.info("Testing on slave address " + str(self.address))
+        if not self.address in locks:
+            locks[self.address] = threading.Lock()
+
+        with locks[self.address]:
+            mbdev = minimalmodbus.Instrument(self.port, slaveaddress=self.address, mode="rtu", close_port_after_each_call=True, debug=False)
+            mbdev.serial.parity = minimalmodbus.serial.PARITY_NONE
+            mbdev.serial.stopbits = serial.STOPBITS_ONE
+            mbdev.serial.baudrate = 9600
+            mbdev.serial.timeout = 0.4 # yes, 400ms is long but the BMS is sometimes really slow in responding, so this seems a good compromize
+            mbdevs[self.address] = mbdev
+ 
+            for n in range(1, RETRYCNT):
+                try:
+                    string = mbdev.read_string(7, 13)
+                    time.sleep(SLPTIME)
+                    result = True
+                    logger.info("found in try " + str(n) + "/" + str(RETRYCNT) + " for " + self.port + "(" + str(self.address) + "): " + string)
+                except Exception as e:
+                    logger.warn("testing failed (" + str(e) + ") " + str(n) + "/" + str(RETRYCNT) + " for " + self.port + "(" + str(self.address) + ")")
+                    continue
+                break
+            
+        return result and self.read_status_data() and self.get_settings() and self.refresh_data()
+        
+    def get_settings(self):
+        self.max_battery_voltage = self.max_cell_voltage * self.cell_count
+        self.min_battery_voltage = self.min_cell_voltage * self.cell_count
+
+        return True
+
+
+    def refresh_data(self):
+        # call all functions that will refresh the battery data.
+        # This will be called for every iteration (1 second)
+        # Return True if success, False for failure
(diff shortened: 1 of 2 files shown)


ramack:
from the CI checking I get the following error:
`Error: Resource not accessible by integration`
and the warning
`Debugging artifacts are unavailable since the 'init' Action failed before it could produce any.`

@Louisvdw or @mr-manuel 
Could you guide me how to solve those?

mr-manuel:
Sure! Thanks for contribution.

Probably you have to enable it in your repository settings under `Actions` -> `General` -> `Actions permissions` select `Allow all actions and reusable workflows`. Check also in your repository settings under `Actions` -> `General` -> `Workflow permissions` if `Read and write permissions` are selected.

mr-manuel:
Please fill out this table for a new BMS. You can do it in your current branch.

https://github.com/Louisvdw/dbus-serialbattery/blob/9eb464790e4706ae3465127335a5fa1471e26039/docs/docs/general/features.md?plain=1#L124-L144

Are there any special instructions needed that the user needs to know to use this BMS? Like what about the two disabled lines? Please add it here. You can do it also in your current branch.

https://github.com/Louisvdw/dbus-serialbattery/blob/9eb464790e4706ae3465127335a5fa1471e26039/docs/docs/general/install.md?plain=1#L122-L124

I will update the docs/contribution guidelines to make things like this more clear and easier. So you have a checklist on what to check.

ramack:
oh, I see that I created the PR against master, should I choose branch dev instead?

ramack:
> Are there any special instructions needed that the user needs to know to use this BMS? Like what about the two disabled lines? Please add it here. You can do it also in your current branch.

In principle I would say there is nothing special to consider ("connecting the RS485 pins from the BMS to a USB-RS485-converter" I would somehow consider evident :wink:)

**For the two disabled lines:** when I started to work on that these were still in utils.py, where the user was more or less intended to do changes. Now I am not sure what to do... the driver supports different modbus addresses (all in 1:255 are possible) and my mid term target would be to to support multiple BMSs on the same USB-converter by different addresses. - In the driver this seems to be not yet supported, so I am not sure what to do until then. (personal note: I have 4 batteries in parallel and so far only connect one to VenusOS, but want to change that, but ideally with one USB-converter only, so I have already assigned different addresses to them for testing)

ramack:
For the last line in the table: "Set battery parameters (DVCC)" I am not sure, which aspect the driver would need to support here, can you check if I did it right?

etc/dbus-serialbattery/bms/heltecmodbus.py
```diff
@@ -0,0 +1,428 @@
+# -*- coding: utf-8 -*-
+# known limitations:
+#   - only BMS variants with 2 cell temperature sensors supported
+#   - some "interesting" datapoints are not read (e. g. registers 52: switch type, 62: bootloader and firmware version)
+#   - SOC not yet resettable from Venus (similary to Daly for support of writing SOC), but modbus write to 120 should be fairly possible)
+
+
+from battery import Protection, Battery, Cell
+from utils import *
+from struct import *
```
mr-manuel, line 10: Please remove wildcard imports

etc/dbus-serialbattery/dbus-serialbattery.py
```diff
@@ -46,6 +47,9 @@
     {"bms": Renogy, "baud": 9600, "address": b"\x30"},
     {"bms": Renogy, "baud": 9600, "address": b"\xF7"},
     {"bms": Seplos, "baud": 19200},
+    # {"bms": HeltecModbus, "baud": 9600, "address": 3},
+    # {"bms": HeltecModbus, "baud": 9600, "address": 2},
+    {"bms": HeltecModbus, "baud": 9600, "address": 1},
```
mr-manuel, line 52: Please sort both alphabetically. You can remove the commented lines, since you can set the address in config file.

mr-manuel:
> oh, I see that I created the PR against master, should I choose branch dev instead?

I changed it to the `dev` branch

> In principle I would say there is nothing special to consider ("connecting the RS485 pins from the BMS to a USB-RS485-converter" I would somehow consider evident 😉)

Please add it to the `config.default.ini` and `utils.py` (that read the value and store it in a variable) under

https://github.com/Louisvdw/dbus-serialbattery/blob/c0056c6ddeb588c4b42decf0df6c63a44490150a/etc/dbus-serialbattery/config.default.ini#L215

I would recommend to add something like:

```ini
; -- HeltecSmartBMS settings
; Set the Modbus address from the adapter
HELLTEC_MODBUS_ADDR = 
```

then in a later moment, then multiple Modbus clients are allowed you can change it to:

```ini
; -- HeltecSmartBMS settings
; Set the Modbus addresses from the adapters
; Separate each address by a comma like: 1, 4, 17, ...
HELLTEC_MODBUS_ADDR = 
```

Please specify in the [docs](https://github.com/Louisvdw/dbus-serialbattery/blob/9eb464790e4706ae3465127335a5fa1471e26039/docs/docs/general/install.md?plain=1#L122-L124) that the Modbus address can be changed though the config file.

If you set in the `config.default.ini` the factory set address then I would not add more. If you leave it blank, then please add also, that after the instalation the user has to add this to the config file.

> For the last line in the table: "Set battery parameters (DVCC)" I am not sure, which aspect the driver would need to support here, can you check if I did it right?

This is when the driver sets CVL, CCL and DCL directly (`yes`) or if it's calculated by the driver (`calc`).

mr-manuel:
I added a checklist to the documentation: https://louisvdw.github.io/dbus-serialbattery/general/supported-bms/#add-by-opening-a-pull-request

ramack:
sorry for the millions of commits, but that should be it so far.

docs/docs/general/install.md
```diff
@@ -122,7 +122,7 @@ Select `2` for `nightly build` and then select the branch you want to install fr
 ### BMS specific settings
 
 * ECS BMS &rarr; https://github.com/Louisvdw/dbus-serialbattery/issues/254#issuecomment-1275924313
-
+* HeltecModbus &rarr; set INVERT_CURRENT_MEASUREMENT = -1 and in case the modbus slave address of the BMS was adjusted from the factory default, configure the slave addresses to query in config.ini:HELTEC_MODBUS_ADDR. As always the battery settings shall be configured in the BMS already via app or computer.
```
mr-manuel, line 125: Do you need only in some cases to invert the current?

ramack: This is a good question. It seems to be left from the beginning and as I think.of it it might be better to place the minus into the code...

mr-manuel: That would have been my next question. If it's fixed inverted, then better invert it in the code.

ramack: thanks for that improvement!

ramack:
Thanks for merging. It's an honor for me, that you consider my code useful also for others.

mr-manuel:
You‘re welcome. Thanks for contributing to the project!

Issue #636, opened 2023-05-12
JBD BMS will not be recognized (reopen of #608)


sloesp:
### Describe the problem

This is a reopen of #608. Now with physical cells.

Hello,
i am using a Jaibaida AP21S002 21S 300A BMS with the RS485/USB Interface. If i use the JBDTools on Windows i get Data from the BMS. With Venus OS on RASP-Pi i see the LEDs on the Interface blinking for both directions (send/receive) but the are not BMS Information at the GUI.

Any Ideas?
Thanks so far.
Slotty

### Driver version

0.14.3

### Venus OS device type

Raspberry Pi

### Venus OS version

V2.92 same with 2.93

### BMS type

Smart BMS (LLT, JBD, Overkill Solar)

### Cell count

8

### Connection type

Serial USB adapter to RS485

### Config file

```ini
original utils.py
```

### Relevant log output

```shell
@400000006453a4fd23a7a5bc INFO:SerialBattery:Testing LltJbd
@400000006453a4fd2a21742c INFO:SerialBattery:Testing Ant
@400000006453a4fd3a038a24 ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4fd3a1600b4 ERROR:SerialBattery:>>> ERROR: Incorrect Data
@400000006453a4fd3a1f99bc INFO:SerialBattery:Testing Daly
@400000006453a4fe0e6c9e8c ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4fe0e7ce684 INFO:SerialBattery:Testing Daly
@400000006453a4fe1e60bdb4 ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4fe1e71154c INFO:SerialBattery:Testing Jkbms
@400000006453a4fe2e63b1bc ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4fe2e741124 INFO:SerialBattery:Testing Lifepower
@400000006453a4ff02c8074c ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4ff02d7c2a4 ERROR:SerialBattery:>>> ERROR: Incorrect Data
@400000006453a4ff02e17aec INFO:SerialBattery:Testing Renogy
@400000006453a4ff12c2f4a4 ERROR:SerialBattery:>>> ERROR: No reply - returning
@400000006453a4ff12d3a22c INFO:SerialBattery:Testing Renogy
[...]

PR #673, opened 2023-05-26
bugfix: LLTJBD BMS ignore non ASCII letters for hardware version

idstein:
#636 primary issue fixed for decoding hardware revision and ignore invalid characters


mr-manuel:
Thanks!

PR #685, opened 2023-06-01
Update reinstall-local.sh: Fixed charge current parameter

hoschult:
Update reinstall-local.sh: Corrected charge current parameter for  initial config.ini


mr-manuel:
Thanks

PR #695, opened 2023-06-07
remove scan for devices

seidler2547:
The scan for devices and check if the BMS to test is present doesn't add value. If the device is not within range (or the MAC is wrong), then the subsequent start_scraping call will either fail or fail to produce usable data. Hence, we can save the scanning and save a lot of time on startup, since the case that the BMS is available and responding is the expected case.


mr-manuel:
Thanks for helping to improve the driver! Maybe you could also take a look at https://github.com/Louisvdw/dbus-serialbattery/issues/693. I do not understand why this is happening.

PR #654, opened 2023-05-20
Limit control voltage to max charge voltage

ogurevich:
in my setup cell voltage still default
i have seen , that control_voltage get sometimes  values above 55.2
Screenshot 2023-05-20 at 19 42 24
Screenshot 2023-05-20 at 19 43 38

.... not sure but it could occurs here. Could you take a look at this change.
thank you.

; --------- Cell Voltages ---------
; Description: Cell min/max voltages which are used to calculate the min/max battery voltage
; Example: 16 cells * 3.45V/cell = 55.2V max charge voltage. 16 cells * 2.90V = 46.4V min discharge voltage
MIN_CELL_VOLTAGE   = 2.900
; Max voltage can seen as absorption voltage
MAX_CELL_VOLTAGE   = 3.450
FLOAT_CELL_VOLTAGE = 3.382
TIME_TO_GO_ENABLE = False

PR #698, opened 2023-06-09
Improve Battery Voltage Handling in Linear Absorption Mode

ogurevich:
This pull request introduces enhancements to the battery voltage management, specifically during the transition to the Linear Absorption mode. The main changes include:

1. Addition of `PENALTY_BUFFER` to the penalty sum calculation: This change ensures that the penalty sum calculation is more accurate, preventing overcharge of any single cell.

2. Resetting `self.max_voltage_start_time` when `voltageSum` falls below the maximum: This change addresses a specific scenario where a single cell's voltage exceeds the maximum while the total battery voltage (`voltageSum`) has not yet reached the maximum. In this case, `voltageSum` can fluctuate around the threshold value just before the start of Linear Absorption mode. By resetting `self.max_voltage_start_time` during a voltage drop, the code can more accurately track the start of the Linear Absorption mode.

These changes aim to improve the accuracy and efficiency of battery voltage management, particularly during the transition to Linear Absorption mode. They should help prevent overcharging of individual cells and ensure a smoother transition to Linear Absorption mode.

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -68,7 +68,9 @@ class Battery(ABC):
         self.max_battery_charge_current = None
         self.max_battery_discharge_current = None
         self.has_settings = 0
-
+        self.max_battery_voltage = None
+        self.min_battery_voltage = None
+        
         self.init_values()
 
         # used to identify a BMS when multiple BMS are connected - planned for future use
@@ -203,7 +205,7 @@ class Battery(ABC):
                 self.manage_charge_voltage_step()
         # on CVCM_ENABLE = False apply max voltage
         else:
-            self.control_voltage = round((utils.MAX_CELL_VOLTAGE * self.cell_count), 3)
+            self.control_voltage = round((self.max_battery_voltage), 3)
             self.charge_mode = "Keep always max voltage"
 
     def manage_charge_voltage_linear(self) -> None:
@@ -217,9 +219,7 @@ class Battery(ABC):
         tDiff = 0
         
         PENALTY_BUFFER = 0.010
-        MAX_BATTERY_VOLTAGE = utils.MAX_CELL_VOLTAGE * self.cell_count
-        MIN_BATTERY_VOLTAGE = utils.MIN_CELL_VOLTAGE * self.cell_count
-
+        
         try:
             if utils.CVCM_ENABLE:
                 # calculate battery sum
@@ -239,7 +239,7 @@ class Battery(ABC):
                 if self.max_voltage_start_time is None:
                     # start timer, if max voltage is reached and cells are balanced
                     if (
-                        MAX_BATTERY_VOLTAGE - utils.VOLTAGE_DROP
+                        self.max_battery_voltage - utils.VOLTAGE_DROP
                         <= voltageSum
                         and voltageDiff
                         <= utils.CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL
@@ -260,7 +260,7 @@ class Battery(ABC):
                     if 300 < tDiff:
                         self.allow_max_voltage = False
                         self.max_voltage_start_time = None
-                    if voltageSum < MAX_BATTERY_VOLTAGE - utils.VOLTAGE_DROP:
+                    if voltageSum < self.max_battery_voltage - utils.VOLTAGE_DROP:
                         self.max_voltage_start_time = None
 
             # INFO: battery will only switch to Absorption, if all cells are balanced.
@@ -278,9 +278,9 @@ class Battery(ABC):
                         min(
                             max(
                                 voltageSum - penaltySum,
-                                MIN_BATTERY_VOLTAGE,
+                                self.min_battery_voltage,
                             ),
-                            MAX_BATTERY_VOLTAGE,
+                            self.max_battery_voltage,
                         ),
                         3,
                     )
@@ -305,7 +305,7 @@ class Battery(ABC):
 
             elif self.allow_max_voltage:
                 self.control_voltage = round(
-                    (MAX_BATTERY_VOLTAGE), 3
+                    (self.max_battery_voltage), 3
                 )
                 self.charge_mode = (
                     "Bulk" if self.max_voltage_start_time is None else "Absorption"
@@ -348,9 +348,7 @@ class Battery(ABC):
 
                 if self.max_voltage_start_time is None:
                     # check if max voltage is reached and start timer to keep max voltage
-                    if (
-                        utils.MAX_CELL_VOLTAGE * self.cell_count
-                    ) - utils.VOLTAGE_DROP <= voltageSum and self.allow_max_voltage:
+                    if self.max_battery_voltage - utils.VOLTAGE_DROP <= voltageSum and self.allow_max_voltage:
                         # example 2
                         self.max_voltage_start_time = time()
 
@@ -377,7 +375,7 @@ class Battery(ABC):
                         pass
 
             if self.allow_max_voltage:
-                self.control_voltage = utils.MAX_CELL_VOLTAGE * self.cell_count
+                self.control_voltage = self.max_battery_voltage
                 self.charge_mode = (
                     "Bulk" if self.max_voltage_start_time is None else "Absorption"
                 )
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -245,8 +245,6 @@ class Battery(ABC):
         penaltySum = 0
         tDiff = 0
 
-        PENALTY_BUFFER = 0.002
-
         try:
             if utils.CVCM_ENABLE:
                 # calculate battery sum
@@ -259,9 +257,7 @@ class Battery(ABC):
                         if voltage > utils.MAX_CELL_VOLTAGE:
                             # foundHighCellVoltage: reset to False is not needed, since it is recalculated every second
                             foundHighCellVoltage = True
-                            penaltySum += (
-                                voltage - utils.MAX_CELL_VOLTAGE + PENALTY_BUFFER
-                            )
+                            penaltySum += voltage - utils.MAX_CELL_VOLTAGE
 
                 voltageDiff = self.get_max_cell_voltage() - self.get_min_cell_voltage()
 
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -241,7 +241,7 @@ class Battery(ABC):
                         <= utils.CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL
                         and self.allow_max_voltage
                     ):
-                        self.max_voltage_start_time = time()
+                        self.max_voltage_start_time = int(time())
 
                     # allow max voltage again, if cells are unbalanced or SoC threshold is reached
                     elif (
@@ -250,7 +250,7 @@ class Battery(ABC):
                     ) and not self.allow_max_voltage:
                         self.allow_max_voltage = True
                 else:
-                    tDiff = time() - self.max_voltage_start_time
+                    tDiff = int(time()) - self.max_voltage_start_time
                     # if utils.MAX_VOLTAGE_TIME_SEC < tDiff:
                     # keep max voltage for 300 more seconds
                     if 300 < tDiff:
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -215,6 +215,8 @@ class Battery(ABC):
         voltageSum = 0
         penaltySum = 0
         tDiff = 0
+        
+        PENALTY_BUFFER = 0.010
 
         try:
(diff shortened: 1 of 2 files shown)


etc/dbus-serialbattery/battery.py
```diff
@@ -215,6 +215,10 @@ def manage_charge_voltage_linear(self) -> None:
         voltageSum = 0
         penaltySum = 0
         tDiff = 0
+        
+        PENALTY_BUFFER = 0.010
+        MAX_BATTERY_VOLTAGE = utils.MAX_CELL_VOLTAGE * self.cell_count
+        MIN_BATTERY_VOLTAGE = utils.MIN_CELL_VOLTAGE * self.cell_count
```
mr-manuel, line 221: Maybe it also would make more sense to make the calculations available for all functions in the battery class and round the calculated value like in line 206. What do you think?

ogurevich:
Thank you for your suggestion. You're absolutely right, it makes more sense to make the calculations available for all functions

Mit freundlichen Grüßen/ best regards

Oleg Gurevich

PGP fingerprint: 38A0 D0CC BD23 1707 B0AF  D158 E9D7 6E3F E74A 0B0C

etc/dbus-serialbattery/battery.py
ogurevich: Thank you for your suggestion. You're absolutely right, it makes more sense to make the calculations available for all functions

mr-manuel:
Please make sure, that the GitHub Actions run in your repository. See last checkmark on this page [Add by opening a pull request](https://louisvdw.github.io/dbus-serialbattery/general/supported-bms#add-by-opening-a-pull-request).

Now you are not calculating the min/max battery voltages anymore?

ogurevich:
> Please make sure, that the GitHub Actions run in your repository. See last checkmark on this page [Add by opening a pull request](https://louisvdw.github.io/dbus-serialbattery/general/supported-bms#add-by-opening-a-pull-request).
> 
> Now you are not calculating the min/max battery voltages anymore?

roger, wilco
jkbms_ble.py fails in action flake8, will fix as well...
thx

etc/dbus-serialbattery/battery.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, port, baud, address):
         self.max_battery_charge_current = None
         self.max_battery_discharge_current = None
         self.has_settings = 0
+        self.max_battery_voltage = None
+        self.min_battery_voltage = None
```
mr-manuel, line 72: Now you are not calculating the min/max battery voltages anymore?

ogurevich: Oops, sorry. This is bug, where is the place where these values can be calculated? At this point is still self.cell_count = None (I have to get used to python) Would you please give a tip on where this can be done? Don't the implementations of the BMS do that?

ogurevich:
Screenshot 2023-06-10 at 13 11 12

etc/dbus-serialbattery/battery.py
mr-manuel: No problem. Everyone has to start somewhen :-) I put it in the `init_values()` function, so you so more changes needed.

ogurevich:
thank you 👍

mr-manuel:
Is there a reason why you changed the penalty buffer from `-0.010` to `+0.010` and then to `+0.002`? Has this to do with your screenshot above? Since this should not be changed it should not be needed to declare a costant for this.

This is because the cell voltage is not exactly on `3.450 V`, if the penalty is triggered. The voltage is always a bit higher, therefore we have to add this value. Before the user could add this on his own, but this made the driver more complex, so I choose a mean value from my observations.

grafik

ogurevich:
@init time the value of self.cell_count = None the calculation of max / min  fails in this case

mr-manuel:
You are right. Then `manage_charge_voltage()` is the best place. Have you time to test it?

ogurevich:
yes I do (testing), now the conditions are quite good. The sun is shining ...

ogurevich:
Screenshot 2023-06-10 at 14 29 31
looks good

ogurevich:
Originally it was like this, if the cell voltage was just slightly above the max value (delta < 0.01), the result was a negative value.
This can accumulate with several cells. Later, when we calculate voltageSum - penaltySum, we are above the max_battery_voltage. We roughly prevented that #654 without looking long for the cause. As a constant, I renamed it so that you can see the meaning of this reduction more clearly. Basically we want to reduce the voltage a bit when a cell is above the max value. From this perspective, with this additional value (penalty_buffer) we can lower the tension a little further (i.e. a + )

ogurevich:
some other explanation: let's see on state of master branch: line 216
```
                        # calculate penalty sum to prevent single cell overcharge by using current cell voltage
                        if voltage > utils.MAX_CELL_VOLTAGE:
                            # foundHighCellVoltage: reset to False is not needed, since it is recalculated every second
                            foundHighCellVoltage = True
                            penaltySum += voltage - utils.MAX_CELL_VOLTAGE - 0.010
```
and setting of value in line 259
```
 # Keep penalty above min battery voltage and below max battery voltage
                    self.control_voltage = round(
                        min(
                            max(
                                voltageSum - penaltySum,
                                utils.MIN_CELL_VOLTAGE * self.cell_count,
                            ),
                            utils.MAX_CELL_VOLTAGE * self.cell_count,
                        ),
                        3,
                    )
```
assume our MAX_CELL_VOLTAGE = 3.45 and MAX_BATTERY_VOLTAGE=55.2

Example cells voltages:
cells = [ 
    3.44, 3.448, 3.448, 3.448, 
    3.448, **3.472, 3.452, 3.452, 
    3.452, 3.452, 3.452, 3.452, 
    3.452, 3.452, 3.452, 3.452** ]

**voltageSumm: 55.224**

the overvoltage of cell looks like:
overvoltageOfCells: [0,0,0,0,
0,0.021999999999999797,**0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797,0.0019999999999997797**]

the overvoltage of cell[5] has 0.021999999999999797 volt
the rest of them only 0.0019 volt

summOvervoltage: 0.04199 volt
we would like to avoid overvoltage and slightly reduce control voltage 
the value of control_voltage should be lower then voltageSumm: 55.224 

reduced controlVoltage: voltageSumm - summOvervoltage = 55.182

but we calculate penaltySumm with -0.01 
penaltyCells: [0,0,0,0,0,0.011999999999999797,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022,-0.00800000000000022]
and penaltyCellsSumm: **-0.06800000000000242**

so we set
self.control_voltage = voltageSum - penaltySum = 55.2919 that is more then we would like to set.
On you graph at 16:40 the some cells get overloaded. The negativ penaltySumm is the cause in this case.

here is some example in JS for this calculation
```
const MAX_CELL_VOLTAGE = 3.45;
// const cells = [ 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45, 3.45 ]
const cells = [ 
    3.44, 3.448, 3.448, 3.448, 
    3.448, 3.472, 3.452, 3.452, 
    3.452, 3.452, 3.452, 3.452, 
    3.452, 3.452, 3.452, 3.452 ];
console.log("cellVoltage: %s", JSON.stringify(cells));
    
const MAX_BATTERY_VOLTAGE = cells.length * MAX_CELL_VOLTAGE;
console.log("MAX_BATTERY_VOLTAGE: %s", MAX_BATTERY_VOLTAGE);

const voltageSumm = cells.reduce( (accumulator, currentValue) => accumulator + currentValue, 0 );    
console.log("voltageSumm: %s", voltageSumm);

const overvoltageCells = cells.map( cellVoltage => cellVoltage > MAX_CELL_VOLTAGE?cellVoltage-MAX_CELL_VOLTAGE: 0);
console.log("overvoltageOfCells: %s", JSON.stringify(overvoltageCells));

const summOvervoltage = overvoltageCells.reduce( (accumulator, cellOvervoltage) => accumulator + cellOvervoltage, 0 ); 
console.log("summOvervoltage: %s", summOvervoltage);
console.log("reduced controlVoltage: voltageSumm - summOvervoltage = %s", voltageSumm - summOvervoltage);

const penaltyCells = overvoltageCells.map( overvoltageCell => overvoltageCell > 0?overvoltageCell - 0.01: 0);
console.log("penaltyCells: %s", JSON.stringify(penaltyCells));

const penaltyCellsSumm = penaltyCells.reduce( (accumulator, cellPenalty) => accumulator + cellPenalty, 0 ); 
console.log("penaltyCellsSumm: %s", penaltyCellsSumm);

const penaltySum = overvoltageCells.reduce( (accumulator, cellOvervoltage) => 
    cellOvervoltage > 0?(accumulator + cellOvervoltage - 0.01):accumulator, 0 ); 
console.log("penaltySum: %s", penaltySum);

console.log(" self.control_voltage = voltageSum - penaltySum = %s", voltageSumm  - penaltySum);
```
I thought that we wanted to use penalty_buffer to lower the voltage a little further, so there must be a "+0.01". That was the reason for changing from - to +

I suggest we drop the penalty_buffer

P.S. on your graph at 15:20 the max charge voltage 62v but starting at 16:47 algorithm increased  charge voltage.

etc/dbus-serialbattery/battery.py
```diff
@@ -272,12 +278,14 @@ def manage_charge_voltage_linear(self) -> None:
...
                         self.allow_max_voltage = True
                 else:
-                    tDiff = time() - self.max_voltage_start_time
+                    tDiff = int(time()) - self.max_voltage_start_time
                     # if utils.MAX_VOLTAGE_TIME_SEC < tDiff:
                     # keep max voltage for 300 more seconds
                     if 300 < tDiff:
                         self.allow_max_voltage = False
                         self.max_voltage_start_time = None
+                    if voltageSum < self.max_battery_voltage - utils.VOLTAGE_DROP:
```
mr-manuel, line 287: Please add a comment for what this check is

etc/dbus-serialbattery/battery.py
```diff
@@ -312,17 +320,20 @@ def manage_charge_voltage_linear(self) -> None:
...
-                self.control_voltage = round(
-                    (utils.MAX_CELL_VOLTAGE * self.cell_count), 3
-                )
+                self.control_voltage = round(self.max_battery_voltage, 3)
                 self.charge_mode = (
-                    "Bulk" if self.max_voltage_start_time is None else "Absorption"
+                    # "Bulk" if self.max_voltage_start_time is None else "Absorption"
+                    "Bulk"
+                    if self.max_voltage_start_time is None
+                    else "Absorption" + "(tDiff: " + str(tDiff) + ")"
```
mr-manuel, line 336: This was only for debugging?

ogurevich: yes i was a "debug", ... will fix back

etc/dbus-serialbattery/battery.py
ogurevich: if battery get almost full charged, the charge_mode toggles back and forth  bulk dynamic and absorption dynamic. Activation of absorption (dynamic) starts timer max_voltage_start_time but
does not reset it wenn battery left absorption (dynamic) and goes to bulk (dynamic) 

This means that if the battery jumps back into absorption (dynamic), the cell will still run with the old value. 

If we left absorption (dynamic) ----> voltageSum < self.max_battery_voltage - utils.VOLTAGE_DROP
we always reset  max_voltage_start_time

This debug output illustrated it.

it means 
with # + str(tDiff) in mode 
```
self.charge_mode = (
                    "Bulk dynamic"
                    # + " (vS: "
                    # + str(round(voltageSum, 2))
                    # + " - pS: "
                    # + str(round(penaltySum, 2))
                    # + ")"
                    if self.max_voltage_start_time is None
                    else "Absorption dynamic"
                    # + "(vS: "
                    # + str(round(voltageSum, 2))
                    # + " tDiff: "
                    # + str(tDiff)
                    # + " pS: "
                    # + str(round(penaltySum, 2))
                    # + ")"
                )
```
.....
Resetting self.max_voltage_start_time when voltageSum falls below the maximum: This change addresses a specific scenario where a single cell's voltage exceeds the maximum while the total battery voltage (voltageSum) has not yet reached the maximum. In this case, voltageSum can fluctuate around the threshold value just before the start of Linear Absorption mode. By resetting self.max_voltage_start_time during a voltage drop, the code can more accurately track the start of the Linear Absorption mode.
.....

etc/dbus-serialbattery/battery.py
ogurevich: fix pushed

mr-manuel:
Thanks for helping to improve the driver!

PR #706, opened 2023-06-12
This changes how strict we are about the accepted return data length for the requests to the BMS

wollew:
Maybe this is enough to fix the problems with grid meters accidentally being recognized as Seplos BMSs.


mr-manuel:
Thanks!

PR #735, opened 2023-06-26
fix proposal for #733

### LINKED ISSUE #733: permanently in absorption mode
### Describe the problem

I have two different installations.
in both I've recently noticed that it no longer comes into float mode. the voltage of the battery is at max_cell_voltage * cells and the maximum cell drift is 0.002V - 0.007V

what can I do?

I first switched to step mode but would like to continue linearly

thanks for the support

### Driver version

newest on dev

### Venus OS device type

Raspberry Pi

### Venus OS version

3.00

### BMS type

Daly Smart BMS

### Cell count

15

### Connection type

Serial USB adapter to RS485

### Config file

```ini
default config, no changes
```

### Relevant log output

```shell
im not have input
```

### Any other information that may be helpful

_No response_

ogurevich:
allow "Measurement" und "Variation Tolerance" to keep in absorption mode within MAX_VOLTAGE_TIME_SEC. Fixes #733

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -68,7 +68,9 @@ class Battery(ABC):
         self.max_battery_charge_current = None
         self.max_battery_discharge_current = None
         self.has_settings = 0
-
+        self.max_battery_voltage = None
+        self.min_battery_voltage = None
+        
         self.init_values()
 
         # used to identify a BMS when multiple BMS are connected - planned for future use
@@ -203,7 +205,7 @@ class Battery(ABC):
                 self.manage_charge_voltage_step()
         # on CVCM_ENABLE = False apply max voltage
         else:
-            self.control_voltage = round((utils.MAX_CELL_VOLTAGE * self.cell_count), 3)
+            self.control_voltage = round((self.max_battery_voltage), 3)
             self.charge_mode = "Keep always max voltage"
 
     def manage_charge_voltage_linear(self) -> None:
@@ -217,9 +219,7 @@ class Battery(ABC):
         tDiff = 0
         
         PENALTY_BUFFER = 0.010
-        MAX_BATTERY_VOLTAGE = utils.MAX_CELL_VOLTAGE * self.cell_count
-        MIN_BATTERY_VOLTAGE = utils.MIN_CELL_VOLTAGE * self.cell_count
-
+        
         try:
             if utils.CVCM_ENABLE:
                 # calculate battery sum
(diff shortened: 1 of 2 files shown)


mr-manuel:
I reopened it and changed the repo from `master` to `dev`.

ogurevich:
> I reopened it and changed the repo from `master` to `dev`.

thank you !

etc/dbus-serialbattery/battery.py
```diff
@@ -250,7 +250,8 @@ def manage_charge_voltage_linear(self) -> None:
         voltageSum = 0
         penaltySum = 0
         tDiff = 0
-
+        # meassurment and variation tolerance in volts
+        measurementToleranceVariation = 0.0215
```
mr-manuel, line 254: I would not use more than 3 decimals. Is there a reason why you chose to use 4?

etc/dbus-serialbattery/battery.py
```diff
@@ -269,7 +270,10 @@ def manage_charge_voltage_linear(self) -> None:
             if self.max_voltage_start_time is None:
                 # start timer, if max voltage is reached and cells are balanced
                 if (
-                    self.max_battery_voltage - utils.VOLTAGE_DROP <= voltageSum
+                    self.max_battery_voltage
+                    - utils.VOLTAGE_DROP
+                    - measurementToleranceVariation
```
mr-manuel, line 275: I don't think that this should be added here, else the problem is only moved to a lower value.

I my opinion the correct should be e.g. start at 55.2, but reset only, if going below 55.1785 (55.2 - `measurementToleranceVariation`).

etc/dbus-serialbattery/battery.py
ogurevich: not really, some value between 0.02 and 0.03 should be Ok, that is not in principle. For me was value 0.022 most optimal. Please set any value of you choice.   (I'm afraid to suggest that as a parameter)

etc/dbus-serialbattery/battery.py
ogurevich: i agree. the first one ist not necessary. you are right.

etc/dbus-serialbattery/battery.py
ogurevich: The optimal value may be related to CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL and also to specific load fluctuations in the entire system.

ogurevich: Example with Value 0.03
Screenshot 2023-06-26 at 22 51 04

With Value 0.0215 it takes some more time to meet conditions (better for battery)
Screenshot 2023-06-26 at 22 50 33

PR #773, opened 2023-08-06
add battery current filtering for Daly BMS

Sam3e2:
As mentioned in #739, Daly BMS struggle at current measurements. I had the same problem and it seems that the BMS samples random values within the current waveform. This fork implements a configurable first order low pass filter to get stable current values. 

I am not sure, how common this problem is. Currently, this fork enables filtering by default.

### DIFF
--- a/etc/dbus-serialbattery/bms/daly.py
+++ b/etc/dbus-serialbattery/bms/daly.py
@@ -230,7 +230,7 @@ class Daly(Battery):
             )
             if crntMinValid < current < crntMaxValid:
                 self.voltage = voltage / 10
-                
+
                 if self.current is None or utils.AVERAGE_CURRENT_TIME_CONSTANT == 0:
                     self.current = current
                 else:
@@ -240,7 +240,7 @@ class Daly(Battery):
                     w = 2 * math.pi * dT / utils.AVERAGE_CURRENT_TIME_CONSTANT
                     a = w / (w + 1)
                     self.current = (1 - a) * self.current + a * current
-                
+
                 self.soc = soc / 10
                 return True
 
--- a/etc/dbus-serialbattery/bms/daly.py
+++ b/etc/dbus-serialbattery/bms/daly.py
@@ -6,6 +6,7 @@ from struct import unpack_from, pack_into
 from time import sleep, time
 from datetime import datetime
 from re import sub
+import math
 
 
 class Daly(Battery):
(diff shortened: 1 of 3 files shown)


mr-manuel:
There is already a global current average in the latest dev branch. Maybe we add an option to display it as main current?

@teefix @transistorgit do you have the same problem?

https://github.com/Louisvdw/dbus-serialbattery/blob/b4b9552ca603683ca62eef7197f6dfb01b4791a9/etc/dbus-serialbattery/battery.py#L86-L87

https://github.com/Louisvdw/dbus-serialbattery/blob/b4b9552ca603683ca62eef7197f6dfb01b4791a9/etc/dbus-serialbattery/dbushelper.py#L625-L629

@Sam3e2 please check that the GitHub Actions run. Here you find how: https://louisvdw.github.io/dbus-serialbattery/general/supported-bms#add-by-opening-a-pull-request

Sam3e2:
I am not sure about the actions. In my repo it ran successfully, but I don't see it in the PR.

If the global averaging is used, it should be configurable: Currently it would average over 150s and result in a very slow changing current value. In case I am not the only one having troubles with a Daly BMS, the averaging should be just long enough for the current values to become stable: In my case, the values are oscillating between few amps and several 10 amps, but the 10 second time constant low passed current matches the current of the inverter approximately. 

I think, since the PR fixes a problem with Daly BMS, it belongs to the driver. The global (long time) averaging could be used on top of that.

mr-manuel:
All fine now with the GitHub Actions.

Ok, let‘s wait for feedback from others. Thanks for your contribution.

transistorgit:
Yes, I had the same problem. 

That's why I already implemented a exponential smoothing filter in daly.py:245 (your dev branch)
😉

Sorry, I'm not at my computer, so its hard to look it up and link it with the phone app. But I think you can find it easily

Will be back from vacation soon

Brgds
Bernd

Sam3e2:
> That's why I already implemented a exponential smoothing filter in daly.py:245 (your dev branch)

True, I haven't found that before. This PR implements the same, but configurable.

Tingeltangelbobi:
Hello, I also have this problem. It occurs when I discharge or charge with the MultiPlus2.
If I use a lab power supply for charging, the values are displayed correctly.
The Daly bms has the latest firmware (BmsGD32E230_309(31_230613_001T)) from Daly installed.
I have also sent a bug report to Daly.

mr-manuel:
There is already a current filtering from @transistorgit therefore I won't merge this.

https://github.com/Louisvdw/dbus-serialbattery/blob/1137b3e4e79bd82de1fb556505c818d95d9f16f8/etc/dbus-serialbattery/bms/daly.py#L244-L251

Sam3e2:
I accept that you don't wanna merge it but I don't think that a hard coded time constant is a good solution as it depends on bms sampling rate and user preferences: The time constant is a trade off between accuracy and speed. 

@transistorgit do you have any plans making the time constant configurable?

mr-manuel:
@Sam3e2 can you then please fork the current `dev` branch, integrate your change instead of the above code and open a new PR?

transistorgit:
> @transistorgit do you have any plans making the time constant configurable?

Maybe I'll do that later in fall when I'm going to work on my system again. But you can also add the configuration yourself, if you need it ealier.

PR #776, opened 2023-08-09
bugfix: Type conversion NoneType to Int

idstein:
TypeError("int() argument must be a string, a bytes-like object or a number, not 'NoneType'") of type <class 'TypeError'> in /opt/victronenergy/dbus-serialbattery/dbushelper.py line #643


mr-manuel:
This should already be fixed. Are you using the latest nightly from the dev branch?

mr-manuel:
Was already fixed, but not merged.

PR #813, opened 2023-09-20
Trigger JK BLE SOC reset when using Step Mode

ArendsM:
@mr-manuel Could you please review the position of the SOC reset trigger while charge voltage control is in step mode? I'm not sure if that is the best position.


mr-manuel:
There where you put the `trigger_soc_reset()` for step mode was not the same place as in linear mode. I moved it to the same place and added a check, that it's only called, if the charging mode is changing from `bulk` or `absorption` to `float`.

PR #868, opened 2023-11-20
Calculation of the SOC based on coloumb-counting

cflenker:
SOC is calculated by integrating the current that flows in and out the battery.
The advantage of doing this in the driver is the ability to define the reset-conditions (current, voltage, debouncing-time).
This solves especially the problem of the famous jkbms that only resets the soc when you reach OVP (and the BMS switches off the charging-fets).


mr-manuel:
Please do not merge with the `master` branch. You have to merge with the `dev` branch.

mr-manuel:
I changed it, please update your branch to be up do date with the `dev` branch.

cflenker:
Dear @mr-manuel,
Thank you very much for your reply! I will update the branch to be up to date with the dev-Branch.

gurkc006:
I just had the same idea today and glad to find the "solution" here in the pull requests. Perfect. I would like to test this new feature, if it is ready for testing. I have also a small request: My JKBMS seems to have a small offset in current measurement. Over time, my SOC gets lower and lower (by around 2-3% per day). In summer this is no problem, because it resets every day, but now my SOC drifts down over time. It would be nice, if in calculation of SOC it would be possible do correct for such an offset in the settings of the calculation. Thanks.

gurkc006:
Maybe also a topic here: Wouldn't it be possible to calc a current-corrected cell voltage for supporting estimation of SOC? What I mean: Has someone tried to assume an internal resistance of cables/cells and correct cell voltage with actual current flowing. This should give a quite smooth cell voltage which is not (or much less) affected by current flowing. It also would be possible to "learn" the internal resistance of a system using the current flucuations at different SOC to get better information, what SOC is. Do you get the idea?

cflenker:
Hello @gurkc006,
I had the same issue. The current measurement of the JKBMS was about 5-8% higher than the measurement of the victron smartshunt. I doublechecked it with a clamp meter and it seems that the current of the smartshunt was quite correct and the jlbms was too high.
But there is a possibility to calibrate the current-measurement of the BMS in the jkbms-app. I just tried this and now smartshunt and bms seem to be much closer together. 
I posted a picture of a day of soc measurement of the smartshunt and the bms-driver in this thread:
https://github.com/Louisvdw/dbus-serialbattery/discussions/869
Accordingly to the higher current measurement of the bms the soc of the bms drops faster than the soc of the smartshunt.
Let's wait if the SOC stays closer together now after the calibration of the bms.
If this works, we can perhaps do without an additional correction in the driver.

gurkc006:
Hi cflenker,
thanks for your input. I already did a calibration right at the beginning when top balancing my battery for the very first time, because I had steady charge current from lab power supply. Now on the Multiplus II it's rather difficult to perform good calibration because current from or to battery is very unstable. Also I suspect, that I don't need a span calibration rather then an offset calibration and this is not possible with JKBMS app as far as I know. I see it, if my Multiplus is in idle, JKBMS thinks to see a small current flowing out of battery (around maybe 1 A). But that's 50 W... So over time this adds up.  So for me it would be nice to see, if a offset/span calibration in the driver would lead to a better SOC calculation. But of course, if it's not necessary for many people I can understand, that you won't implement it in the driver. Maybe I try to fork for myself and try. But I never used github, so I hoped I can "attach" to guys like you, who are already doing gihub stuff :-) Thanks a lot

cflenker:
Hi @gurkc006,
I recorded the current values of the victron bms and the jkbms for about 24h. Each point is the average of 30s.
Here is the result:
image
Based on the regression-line the jkbms is about 3.7% too high.
The bigger problem is the big deviation at low discharging values (-6...0A). Here the jkbms shows much higher values.
That's typical the current that is consumed during the night (0...300W).
I trust the victron smartshunt much more than the jkbms. 
But if you implement a correction curve: I wonder who does the effort to measure out the curve.

mr-manuel:
And the question is, if the curve is the same for all... I have measured the base load on my system and then calibrated the JKBMS on the base load that matches 16 of 24 hours of the day

gurkc006:
Hi cflenker!
That's a great plot. And I would definitly try to produce such a plut for my setup. Maybe then we learn something, if it looks the same or not. Problem is, I have no Smartshunt... At the moment I only have a Fluke clamp current meter without serial interface, so I have to take value manually. But the good thing: In the morning I can have only my 4 MPPTs loading the battery starting from 0A to around 20A when sun is shining - so not at the moment :-( - and I can then take very undisturbed calibration points. Maybe I also can use the wires from battery to first junction to measre voltage drop and calibrate this with the fluke. So then I can use a normal good multimeter to log much more points. This will take some days, but I will definitly do this to see, if I can correct for this offset problem. I think, it would be best, if I just can define a pair of values to produce a calibration curve, so the serialbattery-driver can just linear interpolate between these point-pairs to correct the JKBMS curent values.
Thanks a alot for your work! Great!

gurkc006:
This morning I just had the chance to take some values manually. So here are my results from loading my battery with MPPT this morgning:
jkbms_calib1
As you can see, all points are on a good line. I thought about your data cflenker, and the "problem" is not the slope of your fit, because this should neutralize over time. So if JKBMS sees too high currents for charging and discharging, if you sum up all your (a little bit to high) positive and negative numbers, should still give 0 over time. But the offest of -0.25A is there all the time, so you should have also a drift down to lower values, because JKBMS has this offset for ALL values. And this results to a virtual self discharge rate of around 0.25A*24h = 6Ah/d = 2.1% per day (for a 280Ah battery)...
For me this offset is around 0.32A*24h = 2.7% per day (for a 280Ah battery. This fit'S quite well with the numbers I see, if I don't charge my battery full ervery day. So potentially adding this offset to ALL current values coming from JKBMS should compensate for this offest and the virtual self discharge should be gone. 
Will try to make some more measurements on sunday also for discharging the battery. Have a nice weekend.

cflenker:
Thank you very much for your measurements! Your measurements fit very good to a straight line. That's the right way to do it. My measurements have been taken normal operation and this is a very unsteady state.  Your measurement quality seems to be much better.
Your conclusion, that the offset of the regression-curve (0.25A oder 0.32A) leads to soc-loss per day is only valid if all points fit to a straight line. In your measurement this seems to be the case. 
In my measurement it seems that the low current points are relativly too high. 
Anyway - for the first look it seems to be sensible to do a slope+offset-correction. 
The only point is, that if I disconnect the battery the jkbms shows 0 Amps. WIth the correction it would show 0.25A.

cflenker:
I just added the option to make a current correction with offset and factor.
@gurkc006: If you want to try it, you can install this pre-release from my fork of the repo:
https://github.com/cflenker/dbus-serialbattery/releases/tag/v1.0.20231117dev_chfl231124

cflenker:
The more I think about it, it would perhaps be better to make a separate scaling factor for positiv (charging) and negativ (discharging) current. 
That would ensure, that 0A will stay 0A.

cflenker:
I installed the driver on a friend's system. Here are the results of the last week. I installed the driver with coloumb-counting at late evening of 11/22:
image
I had no correction of the current applied.

gurkc006:
I will install your driver tomorrow, when I'm at home again. Personally I would prefer a table solution for the correction. Something like in the other tables in the driver:
I_JKBMS = -20, -10, -1, 0, 1, 5, 10, 25, 70
I_ref = -19.75, -9.75, -0.75, 0.25, 1.25, 5.25,10.25, 25.25, 70.25
Then just linear interpolate between the values and use the last slope for bigger values than in the table. That would be the most flexible solution and maye it's also the easiest, because you don't have to caluclato gain and offset. Just take measurements at different currents and put it into table. What do you think?

cflenker:
I agree. That would be the most flexible solution. But I would never calibrate it with someting else than 0Amps at 0Amps. When the Battery is disconnected the JKBMS measures 0Amps. And I would want the driver to calculate a rising SOC over time when the battery is disconnected.
I will have a look how the interpolation can be done.
@mr-manuel: What do you think about that? Would you agree to spend the computing-time for that?

cflenker:
I think it is a known fact, that the capacity of a battery depends on the discharge rate. This effect is much smaller at lifepo4 than at lead-batteries, but it is also present. It is called the peukert-exponent.
The smartshunt takes this into account (I calibrated it to 1.01). Theoretical it would be able to calibrate this effect into the correction-curve, but you can't do this with plain current-measurement.
Maybe we should think about taking this into account in a future version. (I did not completely understand how to take Peukert into account during charging).

mr-manuel:
The correction with a table should be the simplest one for the end users. So they can measure the current at different values and compare the value with the BMS provided.

Most of them does not have even a current measurement tool. We try to keep it as simple as possible for the user. The question is also, how does this work, if you have multiple batteries? Is this very CPU intensive? Will a weak GX device also be able to handle that?

cflenker:
Thank you @mr-manuel for your quick reply. 
I will make a proposal for that. I would use the **utils.calcLinearRelationship**-Function.
And I would set the default of the look-up as short as possible:
SOC_CALC_CURRENT_MEASURED = -300, 300
SOC_CALC_CURRENT_REAL = -300, 300
For my understanding it is calculated for every battery separately.

gurkc006:
You can make it disabled as default so only users who want to use this feature can enable it in the ini- file?! Should be no problem as it's only maybe once per second to calculate the correction.

cflenker:
I made a "prerelease" last night: https://github.com/cflenker/dbus-serialbattery/releases/tag/v1.0.20231117dev_chfl231125c
This is based on the official dev-branch from 17.11.2023.
It includes the latest "Calculation of the SOC based on coloumb-counting"-Feature and an alternative method for dynamic CVL I am just working on. Both are switched of by default.

I am just doing a few tests on my system with this v1.0.20231117dev_chfl231125c.

My configuration concerning soc_calc is the following:

SOC_CALCULATION = True
SOC_RESET_CURRENT = 7
SOC_RESET_TIME = 60
SOC_CALC_CURRENT_CORRECTION = True
SOC_CALC_CURRENT_MEASURED = -300.0, 300.0
SOC_CALC_CURRENT_REAL = -289.0, 289.0
SOC_CALC_INIT_VALUE = 25.0

gurkc006:
Good morning!
I installed your "prerelease" yesterday, but didn't reboot until now. I saw, that of course the SOC is lost over every reboot and will be initiated with the SOC_CALC_INIT_VALUE. I thought about that and two ideas came into my mind:
1. Would it be possible, to have a writable variable in the Venus OS menue to "reset" the SOC on demand? So I could remember the last SOC before reboot and reset the driver to that value manually after reboot. 
2. An even more elegant solution would be, to save the last calculated SOC on exiting the driver. I had a look into the source files on github, but I'm not sure, if there is the possibility to do something on exiting of the driver at all. Is there a callback function which will be called, when someone tries to reboot the device? That would be perfect. Just save the last SOC to a SOC.ini file and try to reload that after reboot. 
Sorry for having all these requests, I'm perfectly fine with the current solution, but would like to help to improve further. 
I did some load measurements yesterday evening and will check later if they fit to my calibration curve and then will edit the correction table in the new driver and restart and see what happens :-) Thanks a lot!

mr-manuel:
In the latest nightly from the dev branch I implemented that some values are stored in the dbus on change. You can copy the existing code:

Here to save the value:

https://github.com/Louisvdw/dbus-serialbattery/blob/da73f050c7b61e55e4cb472dc8138d38774efb64/etc/dbus-serialbattery/dbushelper.py#L1053-L1070

Here to restore the value:

https://github.com/Louisvdw/dbus-serialbattery/blob/da73f050c7b61e55e4cb472dc8138d38774efb64/etc/dbus-serialbattery/dbushelper.py#L169-L174

gurkc006:
I just updated my calibration curve for my JKBMS:
jkbms_calib2
So it looks, like my span calibration is 3% off (what's fine for me) and the offset is around -260mA. So I will try to use these coefficients to fine-tune your driver internal SOC calculations.

cflenker:
Hello @mr-manuel ,
Thank you very much for your blueprint to save values in the dbus!
I tried to implement it the same way, but I got a "BMS connection loss"-Error of my CerboGX.
Then I tried to install the latest nighty of the dev-Branch (without my PR) and got the same error.
Did you test it sucessfully?
Sorry, I am not so experienced in debugging.
Thank for your help!

mr-manuel:
Can you open an issue for that? So we don't mess up this PR. I saw no error on my two systems where I tested it.

mr-manuel:
@cflenker please try the new nightly build I merged 1h ago. There where indeed some problems that I did not get on my system.

etc/dbus-serialbattery/config.default.ini
```diff
@@ -228,6 +228,28 @@ TIME_TO_SOC_RECALCULATE_EVERY = 60
...
+;        the Soc is resetted to SOC_CALC_INIT_VALUE nevertheless.
+; SOC_CALCULATION = False: Use SOC reported from BMS (none of the other parameters apply)
+SOC_CALCULATION = False
+SOC_RESET_CURRENT = 7
+SOC_RESET_TIME = 60
+SOC_CALC_CURRENT_CORRECTION = False
+SOC_CALC_CURRENT_MEASURED = -300, 300
+SOC_CALC_CURRENT_REAL = -300, 300
+SOC_CALC_RESET_VALUE_ON_RESTART = False
+SOC_CALC_INIT_VALUE = 25
```
mr-manuel, line 252: Please move this below the section `; -- CVL reset based on SoC option (step mode & linear mode)` so we have similar settings in the same place.

In my eyes this are too much options here. We need to keep it as simple as possible

I would use the existing `MAX_VOLTAGE_TIME_SEC` instead of `SOC_RESET_TIME`.

`SOC_CALC_CURRENT_CORRECTION = False` could be removed. Just check, if `SOC_CALC_CURRENT_MEASURED` and `SOC_CALC_CURRENT_REAL` are the same on driver start.

I would also remove `SOC_CALC_RESET_VALUE_ON_RESTART` and `SOC_CALC_INIT_VALUE`. If no SOC is found in dbus, then use the BMS SOC to start. I think it would make no sense to start from a specific value on every driver restart or device reboot.

What do you think about this?

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -1100,4 +1119,18 @@ def saveBatteryOptions(self) -> bool:
...
+        if abs(self.battery.soc_calc - self.save_charge_details_last["soc_calc"]) > 1.0:
+            self.save_charge_details_last["soc_calc"] = self.battery.soc_calc
+            result = result and self.setSetting(
+                get_bus(),
+                "com.victronenergy.settings",
+                self.path_battery,
+                "SocCalculated",
+                self.battery.soc_calc,
+            )
+            logger.info(
```
mr-manuel, line 1131: I added it as info, so see, if the new changes are working, since my added values do not change very often. You can already change it to debug, if it floods your log.

etc/dbus-serialbattery/config.default.ini
cflenker: HI @mr-manuel, thank you very much for your remark.
I added the Option SOC_CALC_RESET_VALUE_ON_RESTART and SOC_CALC_INIT_VALUE to be able to synchronize with the smartshunt on start. But I am fine with your suggestion too. I agree that it is important to keep it simple.
The other topics are totally fine for me. I will change tomorrow evening.
Thank you!

cflenker: If it ok for you I would like to keep `SOC_RESET_TIME`, because it is typically much smaller (about 60s) than `MAX_VOLTAGE_TIME_SEC`.
I removed 
`SOC_CALC_CURRENT_CORRECTION `
`SOC_CALC_RESET_VALUE_ON_RESTART`
`SOC_CALC_INIT_VALUE`

And I moved the paragraph as suggested.

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -178,6 +178,11 @@ def setup_instance(self):
                             value["SocResetLastReached"]
                         )
 
+                    if "SocCalculated" in value and isinstance(
+                        value["SocCalculated"], float
+                    ):
+                        self.battery.soc_calc = float(value["SocCalculated"])
```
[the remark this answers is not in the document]
cflenker: I put it in alphabetical order.

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -277,6 +283,12 @@ def setup_instance(self):
                 0,
                 0,
             ],
+            "SocCalculated": [
+                self.path_battery + "/SocCalculated",
+                self.battery.soc_calc if self.battery.soc_calc else 0.0,
+                0,
+                0,
+            ],
```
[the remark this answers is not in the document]
cflenker: I put it in alphabetical order.

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -1100,4 +1119,18 @@ def saveBatteryOptions(self) -> bool:
...
+                "com.victronenergy.settings",
+                self.path_battery,
+                "SocCalculated",
+                self.battery.soc_calc,
+            )
+            logger.info(
+                f"Saved SocCalculated. Before {self.save_charge_details_last['soc_calc']}, ",
+                +f"after {self.battery.soc_calc}",
+            )
+
```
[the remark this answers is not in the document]
cflenker: I put it in alphabetical order.

etc/dbus-serialbattery/dbushelper.py
cflenker: I changed it to debug.

cflenker:
Hello @mr-manuel,
I fixed the issues you reported.
If you have any more remarks, please let me know.
From my side this PR is ready for review.
Thanks a lot!

etc/dbus-serialbattery/dbushelper.py
mr-manuel, line 1: There is a `,` at the line end. You copied my typo :-)

mr-manuel:
> I agree. That would be the most flexible solution. But I would never calibrate it with someting else than 0Amps at 0Amps.

In this case something like that, could be used:

~~~
-300, 0, 0, 300
-298, -0.25, 0.25, 298
~~~


@cflenker all fine, not needed, I can do a squash merge :-)

gurkc006:
Hi, 
I lost track a little bit. What's the current status of the soc calculation? Which is the version I should try? Is it your's cflenker? Is this already with storing the soc values in dbus for surviving reboots? Or should I test louisvdw:dev? I'm not so familiar with all this... Thanks a lot for all your fantastic work!

gurkc006:
Installed this version two days ago, had a litle trouble with the installation because I didn't find a way to get the proper venus-data.tar.gz file. Had to build one on my own, but finally it seemed to work. Driver is running now, but nearly no sunlight at all in these days in germany... So only very little variation in SOC as my battery rests on 60% at the moment. Thanks for the great work!

mr-manuel:
Did anyone track the CPU usage before and after this commit? That would be interesting.

gurkc006:
How do I check? htop? Nothing to see. But haven't checked before.

mr-manuel:
If you have an InfluxDB and Grafana you can run Telegraf with this version and the normal version.

etc/dbus-serialbattery/dbushelper.py
```diff
@@ -171,6 +174,12 @@ def setup_instance(self):
                             value["MaxVoltageStartTime"]
                         )
 
+                    if "SocCalc" in value:
+                        self.battery.soc_calc = float(value["SocCalc"])
+                        logger.info(f"Soc_calc read from dbus: {int(self.battery.soc_calc)}")
+                    else:
+                        logger.info(f"Soc_calc not found in dbus")
+
```
mr-manuel, line 182: I added a check before this lines get executed:

```python
                    # load SOC from dbus only if SOC_CALCULATION is enabled
                    if utils.SOC_CALCULATION:
                        if "SocCalc" in value:
                            self.battery.soc_calc = float(value["SocCalc"])
                            logger.info(
                                f"Soc_calc read from dbus: {int(self.battery.soc_calc)}"
                            )
                        else:
                            logger.info("Soc_calc not found in dbus")
```

mr-manuel:
@cflenker relooking at the values after a bit of time the naming is a somehow not very clear.

`SOC_CALC_CURRENT_MEASURED` is this the value measured by the user or by the BMS?
`SOC_CALC_CURRENT_REAL` is this the real value from the BMS or the value measured by the user?

I will rename
`SOC_CALC_CURRENT_MEASURED` to `SOC_CALC_CURRENT_REPORTED_BY_BMS`
and
`SOC_CALC_CURRENT_REAL` to `SOC_CALC_CURRENT_MEASURED_BY_USER`
after you confirm me, that this is correct.

cflenker:
Hello @mr-manuel , that is correct. Your proposal is perfect an very clear! Thank you!

cflenker:
Thanks a lot for merging this PR!

mr-manuel:
I tested the changes a bit further and made some little improvements. One thing is that you are currently saving the `SocCalc` as `int` to the dbus on change. Since you are calculating the `SocCalc` from `self.soc_calc_capacity_remain` would it maybe make more sense to save this value to the dbus instead? The bigger the battery is the more inaccurate the soc is.

Regardless of this, I changed the code in order that the soc has two decimals, so there is a smoother graph.

cflenker:
Ohh, I did not know that you are reading the soc directly from the dbus. Then integer is surely not precise enough. I thought dbus is only to preserve the soc for a reset. Please feel free to change whatever you want. You are much experienced than I am with the driver. Thank you very much for your enhancements!

mr-manuel:
I merged my changes. I would be glad, if you could also test the current `nightly` of the `dev` branch.

cflenker:
I have some problems with the nightly. Please see https://github.com/Louisvdw/dbus-serialbattery/issues/898

mr-manuel:
The new JKBMS do this and the smartshunt does not see the cell voltages.

> The reset should only be done if the lowerst cell drops below a voltage limit at low current for a certain time. 

What if my battery is nearly depleeted and then a bigger load turns on and the whole system crashes because the BMS disconnect the battery? In this case I don't think that it has to be at this voltage for a certain amount of time. Once reached lowest disconnect voltage then it should be se to 0 %. Another thing would be, if you are using a voltage above the BMS under voltage protection. What would be the better value to use in your opinion?
On electric cars the 0% SOC is always above the BMS disconnect voltage from what I know.

Vitalic66:
Hi, thanks for your work, just installed dev driver, because I also have big differences between Victron shunt SOC an JK SOC.  Thinking about this solution, might give it a try. But why do you use -300,0,0,300? My JK can only handle up to 200A.

Another idea, isn‘t it possible to use the SOC meassured and calculated from the Smartshunt instead? So if it is allready installed why not using this SOC directly? 

So using the charge and discharge logic from the driver with the SOC reported from the Victron smartshunt.

Honusnap:
> The new JKBMS do this and the smartshunt does not see the cell voltages.
> 
> > The reset should only be done if the lowerst cell drops below a voltage limit at low current for a certain time.
> 
> What if my battery is nearly depleeted and then a bigger load turns on and the whole system crashes because the BMS disconnect the battery? In this case I don't think that it has to be at this voltage for a certain amount of time. Once reached lowest disconnect voltage then it should be se to 0 %. Another thing would be, if you are using a voltage above the BMS under voltage protection. What would be the better value to use in your opinion? On electric cars the 0% SOC is always above the BMS disconnect voltage from what I know.

You can't really set the "0% SOC" on a cell reaching a minimum voltage setting, as you said a high load at 5% SOC could maybe let the weakest cell drop under this voltage and set the SOC to 0%. Maybe a combination of total battery voltage and current, like 45V and under one Amp. => We are at 0% SOC usable.
Like you said, we... at least most people, will use their battery in a safe zone, i'm using mine between 10% and 95% REAL SOC. Now, to reset the coulomb counting we need to reach those limits( 10% and 95% REAL SOC), a 10kWh battery would then only have 85% usefull capacity (in my case) and those 85% are now the 0%-100% displayed SOC.

lifepo4-voltage-chart 1

Honusnap:
> @mr-manuel : calibrating the soc at 0% is not very common. Most of the bms and also the smartshunt does not do this (afaik). But I do not see a reason why not. The reset should only be done if the lowerst cell drops below a voltage limit at low current for a certain time. If you want me to make a proposal, just tell me. But I am on a business-trip till the end of next week.

In your config.ini there are 2 values... is this config valid  : 

SOC_CALC_CURRENT_MEASURED = -200, -100, -50, -25, -10, -5, 0, 5, 10, 25, 50, 100, 200
SOC_CALC_CURRENT_REAL = -200, -100, -50, -25, -10, -5, 0, 5, 10, 25, 50, 100, 200

Can we use decimal numbers : 0.5 ?

mr-manuel:
Your config is valid for an example, but would make no sense for a real config.

Yes you can use decimal numbers, as you see in the `config.default.ini`:

https://github.com/Louisvdw/dbus-serialbattery/blob/ce2df1a3fed55ecabc8c5722d742f9c0f27aad6b/etc/dbus-serialbattery/config.default.ini#L170-L172

Vitalic66:
So to have valid values we'd need a USB DC Amp data logger to have some values we can feed to the config.ini? Any good recommendations?

Honusnap:
Here is what i got after some experiments :

Huuh-17:
Hello, i need some help to create a code for a circuit with current sensor arduino lithium ion battery usb charger and lcd monitor i need to display the  SOC using the coloumb counting method

PR #882, opened 2023-11-29
Enhancement of CVL-controller 

cflenker:
A big advantage of serialbattery-driver compared to a victron-system with a smartshun-system is, that the serialbus-driver sees every cell voltage and can react on bad balanced batteries.

Especially at the end of charging (close to 100%) the charging-current is usually controlled indirect by CVL (Charge Voltage Limitation).
Furthermore in Victron Systems with MPPTs and DC-Feed-In activated CVL is the only way to reduce charging current, because CCL (charge current limitation does only apply to Multis, but not to MPPTs.

So it is a very good feature of the driver, that CVL can be reduced if a single cell runs over MAX_CELL_VOLTAGE.
This is implemented today with a "penaltySum"-Method that is in fact a P-Controller.

I think an I-Controller fits better to the requirements of CVL-Controlling.
This is implemented with this PR.
With the switch `CVL_ICONTROLLER_MODE` the PenaltySum (`CVL_ICONTROLLER_MODE`=false) or a I-Controller (`CVL_ICONTROLLER_MODE`=True) can be choosen.

I made a test with each controller. With a power-supply I charged a single cell (cell no3) for 1.5h with 2Amps. So I got the cell3 about 3Ah higher than the rest of the 16s-Pack.
image

Then I charged the battery-pack to 100%. And the balancer got it right again.

Here is the result:
 
image

An I-Controller provides a calmer CVL and so gives the victron system time to adjust the charging-current to meet the CVL.

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -360,20 +360,23 @@ class Battery(ABC):
                 ):
                     self.max_voltage_start_time = None
 
-            if self.control_voltage:
-                controlvoltage = self.control_voltage - (
-                    (
-                        self.get_max_cell_voltage()
-                        - utils.MAX_CELL_VOLTAGE
-                        - utils.CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL
+            if utils.CVL_ICONTROLLER_MODE:
+                if self.control_voltage:
+                    controlvoltage = self.control_voltage - (
+                        (
+                            self.get_max_cell_voltage()
+                            - utils.MAX_CELL_VOLTAGE
+                            - utils.CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL
+                        )
+                        * utils.CVL_ICONTROLLER_FACTOR
                     )
-                    / 6.0
+                else:
+                    controlvoltage = utils.MAX_CELL_VOLTAGE * self.cell_count
+
+                controlvoltage = min(
+                    max(controlvoltage, self.min_battery_voltage),
+                    self.max_battery_voltage,
                 )
-            else:
-                controlvoltage = utils.MAX_CELL_VOLTAGE * self.cell_count
-            controlvoltage = min(
-                max(controlvoltage, self.min_battery_voltage), self.max_battery_voltage
-            )
 
             # INFO: battery will only switch to Absorption, if all cells are balanced.
             #       Reach MAX_CELL_VOLTAGE * cell count if they are all balanced.
@@ -389,8 +392,10 @@ class Battery(ABC):
                     ),
                     3,
                 )
-                # self.set_cvl_linear(control_voltage)
-                self.control_voltage = controlvoltage
+                if utils.CVL_ICONTROLLER_MODE:
+                    self.control_voltage = controlvoltage
+                else:
+                    self.set_cvl_linear(control_voltage)
 
                 self.charge_mode = (
                     "Bulk dynamic"
@@ -402,8 +407,10 @@ class Battery(ABC):
                     self.charge_mode += " & SoC Reset"
 
             elif self.allow_max_voltage:
-                # self.control_voltage = round(self.max_battery_voltage, 3)
-                self.control_voltage = controlvoltage
+                if utils.CVL_ICONTROLLER_MODE:
+                    self.control_voltage = controlvoltage
+                else:
+                    self.control_voltage = round(self.max_battery_voltage, 3)
 
                 self.charge_mode = (
                     "Bulk" if self.max_voltage_start_time is None else "Absorption"
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -288,6 +288,7 @@ class Battery(ABC):
         voltageSum = 0
         penaltySum = 0
         tDiff = 0
+        controlvoltage = 0
         current_time = int(time())
 
         # meassurment and variation tolerance in volts
@@ -359,6 +360,21 @@ class Battery(ABC):
                 ):
                     self.max_voltage_start_time = None
 
+            if self.control_voltage:
+                controlvoltage = self.control_voltage - (
+                    (
+                        self.get_max_cell_voltage()
+                        - utils.MAX_CELL_VOLTAGE
+                        - utils.CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL
+                    )
+                    / 6.0
+                )
+            else:
+                controlvoltage = utils.MAX_CELL_VOLTAGE * self.cell_count
+            controlvoltage = min(
+                max(controlvoltage, self.min_battery_voltage), self.max_battery_voltage
+            )
+
             # INFO: battery will only switch to Absorption, if all cells are balanced.
             #       Reach MAX_CELL_VOLTAGE * cell count if they are all balanced.
             if foundHighCellVoltage and self.allow_max_voltage:
@@ -373,7 +389,8 @@ class Battery(ABC):
                     ),
                     3,
                 )
-                self.set_cvl_linear(control_voltage)
+                # self.set_cvl_linear(control_voltage)
+                self.control_voltage = controlvoltage
 
                 self.charge_mode = (
                     "Bulk dynamic"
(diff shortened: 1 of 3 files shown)


mr-manuel:
Unfortunately I have no possibility to test this now in winter, but based on the graph it looks very good! If many can test this, I would remove the current method with the penalty voltages.

codeworkx:
I have been using it for about a week and have not noticed anything negative so far. Thanks for your work!

cflenker:
Hello @codeworkx, thank you a lot for your positiv reply. I am also using it for some time now on 2 Systems. Of cause i got only to 100% by planned charging in December. But for this I always see a much smoother behaviour.

mr-manuel:
This PR is now merged and will replace the penalty sum voltage after a few months of testing. In the meanwhile it will be available as option.

cflenker:
Hello @mr-manuel , thanks a lot. Happy to read this!

mr-manuel:
Thank you for your contribution!

mr-manuel:
@cflenker now I tested it with loading the battery from the grid. The battery never reached the `SOC_RESET_VOLTAGE`, but only `MAX_CELL_VOLTAGE`. This is done in this lines:

https://github.com/Louisvdw/dbus-serialbattery/blob/609342638c7d4958f7ff6cf5517c5aa62c22f21c/etc/dbus-serialbattery/battery.py#L269-L278

Checking your changes you did not consider that. Could you use `self.max_battery_voltage` or use the same logic as above?

The curve looks very clean! Great work!

grafik

cflenker:
Hello @mr-manuel, sorry. You are right. I did not consider that. I will make a proposal to fix it tonight or tomorrow  night. Thank you.

mr-manuel:
I already fixed it ;-)

cflenker:
Thank you very much!!!

Becker884:
with CVL_ICONTROLLER_MODE = True, will never be switched to float ?
I tried today, after successfull absorption to <10mV CVL stays at 3,43V (61,8V):
grafik

mdkeil:
Same to me.. 4h absorb with <10mv celldiff. I will switch back to `CVL_ICONTROLLER_MODE = False` for Fürther testing.

cflenker:
Hello @mdkeil, hello @Becker884,
I do not think it has to do with the type of cvl-controller. 
Maybe it has to do with: https://github.com/Louisvdw/dbus-serialbattery/pull/895
Please try to set CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_TIME_RESTART to a higher value. Maybe 0.030V.

mdkeil:
I will try it out.

mdkeil:
Problem solved..  I had to setup a little voltagedrop, because with linear limitation enabled I never reached the maximum charge voltage.

Becker884:
can you show me please what you set up?

mdkeil:
my config:

```
LOGGING = INFO

MAX_BATTERY_CHARGE_CURRENT = 100
MAX_BATTERY_DISCHARGE_CURRENT = 100

MIN_CELL_VOLTAGE   = 2.900
MAX_CELL_VOLTAGE   = 3.450
FLOAT_CELL_VOLTAGE = 3.375

VOLTAGE_DROP = 0.025

SOC_RESET_VOLTAGE = 3.500
SOC_RESET_AFTER_DAYS = 15

CVL_ICONTROLLER_MODE = True
LINEAR_LIMITATION_ENABLE = True
CVCM_ENABLE = True

CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_TIME_RESTART = 0.015
CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL = 0.010
;MAX_VOLTAGE_TIME_SEC = 1800

BMS_TYPE = Seplos

; Exclute this serial devices from the driver startup
; Example: /dev/ttyUSB2, /dev/ttyUSB4
EXCLUDED_DEVICES = /dev/ttyUSB0

; Charge current control management enable (True/False).
CCCM_SOC_ENABLE = True
; Discharge current control management enable (True/False).
DCCM_SOC_ENABLE = True

CC_SOC_LIMIT1 = 98
CC_SOC_LIMIT2 = 95
CC_SOC_LIMIT3 = 91

CC_CURRENT_LIMIT1_FRACTION = 0.15
CC_CURRENT_LIMIT2_FRACTION = 0.35
CC_CURRENT_LIMIT3_FRACTION = 0.50
```

mr-manuel:
@cflenker for the purpose of completion could you explan what happens, if the value `CVL_ICONTROLLER_FACTOR` is changed to a bigger/smaller value?

Albarge:
> @cflenker for the purpose of completion could you explan what happens, if the value `CVL_ICONTROLLER_FACTOR` is changed to a bigger/smaller value?

Yeah, I'd like to know this too, as it would be good to understand what is a slower/ faster/ or higher / lower ratio of voltage limitation.

Also, testing 1.3.20240417dev today (I have 2 runner cells) I see the ICONTROLLER lower the CVL from 28v to 27.2v on one battery, but this isn't picked up by batteryaggregator which still shows 28v, so the system still calls for 28v across the three batteries. - Maybe a batteryaggregator problem though?

The CVL keeps decreasing on Battery 1 whilst cell voltage differential also decreases which seems counter intuitive, unless it's still trying to lower the stsem voltage but that's not being aggregated back to Venus:
icontroller

Battery 1:
Screenshot_20240418-175418

BatteryAggregator:
Screenshot_20240418-175448

cflenker:
Hello,
I'll try to explain the behaviour of `CVL_ICONTROLLER_FACTOR`. Most of it is already explained in `config.default.ini`:
```
;   2. I-Controller (CVL_ICONTROLLER_MODE = True)
;      An I-Controller tries to control the voltage of the highest cell to MAX_CELL_VOLTAGE + CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL.
;      (for example 3.45V+0.01V =3.46V). If the voltage of the highest cell is above this level, CVL is reduced. If the voltage is below, CVL is
;      increased until cellcount*MAX_CELL_VOLTAGE.
;      An I-Part of 0.2 V/Vs (CVL_ICONTROLLER_FACTOR) has proved to be a stable and fast controlling-behaviour.
;      This method is not as fast as the penalty_sum-Method but usually smoother and more stable against toggeling and has no stationary deviation.
```

So the CVL_ICONTROLLER_FACTOR controls how fast the CVL is lowered if one cell goes above `MAX_CELL_VOLTAGE + CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL`.
For example: If `MAX_CELL_VOLTAGE`=3.45V and `CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL`=0.01V and the actual highest cell voltage is 3.56V. Then the deviation is 0.1V. The deviation is multiplied with `CVL_ICONTROLLER_FACTOR`=0.2 V/Vs. So the CVL is reduced 0.02V per second.

cflenker:
This means:
* The higher the value of `CVL_ICONTROLLER_FACTOR` the faster CVL is reduced with a given cell-voltage-overshoot.
* Do not choose too high values for `CVL_ICONTROLLER_FACTOR`: this could make the controlling instable.

Albarge:
Thanks, that makes it much clearer, so in your scenario above `CVL_ICONTROLLER_FACTOR` = 0.4 will reduce the voltage 0.4 V/second.

 I'll just have to test another driver release to see why it didn't aggregate with multiple batteries, and maybe test different factors to catch my runner cells early.

cflenker:
With CVL_ICONTROLLER_FACTOR = 0.4 and a cell voltage overshoot of 0.1V the CVL will reduced 0.4*0.1 = 0.04V per Second.
Take care: I think a bug crept in with a cleanup-commit: https://github.com/Louisvdw/dbus-serialbattery/issues/1038
I think this bug is fixed in the latest version of @mr-manuel repo. I am just about to test this latest nightly.

mr-manuel:
I made the experience that the I-Controller works not so good when you are off grid with a PV inverter. In this case multiple cells can have a bigger voltage then max voltage and the I-Controller does not correct this and the mess starts. In this case the P-Controller (penalty sum) works better. I will add this node in the confile file.

PR #895, opened 2023-12-21
Condition for transition to float

cflenker:
I often observe the following behaviour:
When reaching MAX_CELL_VOLTAGE and the current is dropping towards zero the cells with higher voltage come down and the cells with lower voltage come up, then crossing each other and drifting away again before the balancer brings them finally all together.
At the point of time when they are crossing (celldiff<`CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL`) the battery pack seems to be balanced an the timer of `MAX_VOLTAGE_TIME_SEC` starts. So after `MAX_VOLTAGE_TIME_SEC` the control_voltage is lowered to FLOAT-Voltage no matter how well they are together at this point of time (as long as they are not more than `CELL_VOLTAGE_DIFF_TO_RESET_VOLTAGE_LIMIT` away from each other). 
In reality the battery is not completely balanced and MAX_VOLTAGE is stopped to early:
image

My proposal would be to introduce one more threshold for cell-deviation that causes MAX_VOLTAGE_TIME_SEC to restart if celldiff gets above this threshold:
image

So celldiff has to be smaller than `CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_TIME_RESTART` for `MAX_VOLTAGE_TIME_SEC` to start the transition to float.

As default value `CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_TIME_RESTART` can be choosen much bigger than `CELL_VOLTAGE_DIFF_KEEP_MAX_VOLTAGE_UNTIL` so this change has no impact for existing configurations and to prevent that some configurations (where the celldiff has a lot of noise) would never change to float.

For my configuration it has a very positiv effect:
image


mr-manuel:
Nice catch :-)

cflenker:
Thanks a lot for merging this commit!

mr-manuel:
Thank you for this great improvements :-)

Issue #915, opened 2024-01-11
JK BLE blocking without updating data. 


cupertinomiranda:
### Describe the bug

Hi everyone,

First of all, thanks for this wonderful project. It is awesome.
I am opening a bug which I wish to resolve, if advised how to get to debug it.

What happens is that dbus-serialbattery hangs without keeping all the data provided to VRM and VenusOS static.
What I would expect would be for it to restart automatically through watchdog, but no. :-(
System assumes nothing is wrong and data is good. Once I had the battery drop "0%" (until BMS disconnected it). Still this is a problem since the discharge current is probably being bypassed by the Multiplus.
Restarting dbus-serialbattery solves the block.

I am currently using the v1.0.20240102beta tag and the same happened again today. Previously was using master branch.

Looking forward to your comments and advise on how to properly find the bug.

![image](https://github.com/Louisvdw/dbus-serialbattery/assets/10813002/8caf93db-f917-4bfb-9711-95d0285b3676)

### How to reproduce

Just wait !

### Expected behavior

Not to block !

### Driver version

v1.0.20240102beta

### Venus OS device type

Raspberry Pi 3

### Venus OS version

v3.13

### BMS type

JKBMS / Heltec BMS

### Cell count

16

### Battery count

1

### Connection type

Bluetooth

### Config file

```ini
[DEFAULT]

; --------- Set logging level ---------
; ERROR: Only errors are logged
; WARNING: Errors and warnings are logged
; INFO: Errors, warnings and info messages are logged
; DEBUG: Errors, warnings, info and debug messages are logged
LOGGING = INFO


; --------- Battery Current limits ---------
MAX_BATTERY_CHARGE_CURRENT    = 100.0
MAX_BATTERY_DISCHARGE_CURRENT = 100.0


; --------- Cell Voltages ---------
; Description: Cell min/max voltages which are used to calculate the min/max battery voltage
; Example: 16 cells * 3.45V/cell = 55.2V max charge voltage. 16 cells * 2.90V = 46.4V min discharge voltage
MIN_CELL_VOLTAGE   = 2.900
; Max voltage (can seen as absorption voltage)
MAX_CELL_VOLTAGE   = 3.450
; Float voltage (can be seen as resting voltage)
FLOAT_CELL_VOLTAGE = 3.375


; --------- SOC reset voltage ---------
; Description: May be needed to reset the SoC to 100% once in a while for some BMS, because of SoC drift.
[...]

Issue #720, opened 2023-06-17
Driver gets unresponsive, if connection is lost on Bluetooth


mr-manuel:
### Describe the bug

See https://github.com/Louisvdw/dbus-serialbattery/pull/689#issuecomment-1592977482

### How to reproduce

1. Use a BMS that uses `callback` instead of `polling` (e.g. JKBMS_BLE)
2. Disconnect the BMS physically

### Expected behavior

Driver should restart

### Driver version

v1.0.20230617beta

### Venus OS device type

Raspberry Pi

### Venus OS version

v3.10~8

### BMS type

JKBMS / Heltec BMS

### Cell count

18

### Connection type

Bluetooth

### Config file

```ini
Not needed
```

### Relevant log output

```shell
2023-06-15 12:41:45.074095500 INFO:SerialBattery:processing frame with battery cell info
2023-06-15 12:41:45.080366500 INFO:SerialBattery:--> DBusHelper: publish_battery()
2023-06-15 12:41:45.081041500 INFO:SerialBattery:--> JKBMS_BLE: refresh_data()
2023-06-15 12:41:45.081637500 INFO:SerialBattery:--> JKBMS_BRN: get_status()
2023-06-15 12:41:45.613783500 INFO:SerialBattery:processing frame with battery cell info
2023-06-15 12:41:45.620835500 INFO:SerialBattery:--> DBusHelper: publish_battery()
2023-06-15 12:41:45.621594500 INFO:SerialBattery:--> JKBMS_BLE: refresh_data()
2023-06-15 12:41:45.622258500 INFO:SerialBattery:--> JKBMS_BRN: get_status()
2023-06-15 12:41:46.154417500 INFO:SerialBattery:processing frame with battery cell info
2023-06-15 12:41:46.160787500 INFO:SerialBattery:--> DBusHelper: publish_battery()
2023-06-15 12:41:46.161481500 INFO:SerialBattery:--> JKBMS_BLE: refresh_data()
2023-06-15 12:41:46.162093500 INFO:SerialBattery:--> JKBMS_BRN: get_status()
2023-06-15 12:41:46.709338500 INFO:SerialBattery:processing frame with battery cell info
2023-06-15 12:41:46.720693500 INFO:SerialBattery:--> DBusHelper: publish_battery()
2023-06-15 12:41:46.721490500 INFO:SerialBattery:--> JKBMS_BLE: refresh_data()
[...]

PR #941, opened 2024-01-27
Fix JK BMS connection restart when bluetooth fails.

cupertinomiranda:
This fix installs a new thread to monitor the state of the original scraping thread.
If scraping thread dies, it verifies that it did not because the scraping was intentionally stopped by calling stop_scrapping. When restarting the scrapper, it first calls the bluetooth reset lambda function that was passed in the class contructor, such that bluetooth is ready to make a proper connection.

This problem was reported by me in issue #915 which got closed since it was a duplicate of #720.

Looking forward to your comments !

### DIFF
--- a/etc/dbus-serialbattery/bms/jkbms_ble.py
+++ b/etc/dbus-serialbattery/bms/jkbms_ble.py
@@ -20,7 +20,7 @@ class Jkbms_Ble(Battery):
         super(Jkbms_Ble, self).__init__(address.replace(":", "").lower(), baud, address)
         self.address = address
         self.type = self.BATTERYTYPE
-        self.jk = Jkbms_Brn(address)
+        self.jk = Jkbms_Brn(address, lambda: self.reset_bluetooth())
         self.unique_identifier_tmp = ""
 
         logger.info("Init of Jkbms_Ble at " + address)
--- a/etc/dbus-serialbattery/bms/jkbms_brn.py
+++ b/etc/dbus-serialbattery/bms/jkbms_brn.py
@@ -139,9 +139,12 @@ class Jkbms_Brn:
     # translate info placeholder, since it depends on the bms_max_cell_count
     translate_cell_info = []
 
-    def __init__(self, addr):
+    def __init__(self, addr, reset_bt_callback = None):
         self.address = addr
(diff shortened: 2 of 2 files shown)


mr-manuel:
Thank you very much for contributing! Did you already test it? A warning message when the treat crashed would be useful :-)

cupertinomiranda:
> Thank you very much for contributing! 

Thank you for contributing the project ! :smile: 

> Did you already test it?

I have force tested it by causing an explicit exit from the thread. It did reboot the thread properly in that case.
I am currently using it to make sure that it restarts the thread in case the problem reported in #915 happens.
It would be nice to have the confirmation from issue #720 creators as well, to check if it solves that problem too.

> A warning message when the treat crashed would be useful :-)

The original code already prints a message for both normal and exception thread exits.

mr-manuel:
You think it would be possible to integrate your fix globally that it works for all BLE drivers?

cupertinomiranda:
Very likely, I did only debug the problem in the context of the JK BMS. However, if other BLE implementations have similar designs it is likely they somehow suffer from the same problem.
I will attempt to give it a look and if I find some more generic solution, I will make another pull request.

mr-manuel:
Best would be, if we add a thread here

https://github.com/Louisvdw/dbus-serialbattery/blob/5e025e66ff4887ee5ce3665789e5d2a14e63ec23/etc/dbus-serialbattery/dbus-serialbattery.py#L226-L229

that works similar as this function, if no data is fetched

https://github.com/Louisvdw/dbus-serialbattery/blob/5e025e66ff4887ee5ce3665789e5d2a14e63ec23/etc/dbus-serialbattery/dbushelper.py#L626-L691

cupertinomiranda:
I am working on a general approach to better control driver monitoring for those that need it.
I will define a is_active method in Battery to default to true and move the monitoring to the Battery itself.
If something goes wrong with the driver is_active will return false and the main process should reinitialize the driver.

What do you think ?

mr-manuel:
What if the driver if the bms is stuck and therefore can't update Is_active to false? This is what the current issue is. Therefore we need only a monitoring for the non polling drivers, since for the polling drivers there is already a check.

Issue #950, opened 2024-02-04
Jkbms_Can: Cell count: 1 | Cells populated: 0 && TypeError: '>' not supported between instances of 'float' and 'NoneType'


p0l0us:
### Describe the bug

Hi, I'm trying to run my Jk bms over CAN bus using this project. When I setup everything and make running, the bms entry is appearing and disappearing from the VenusOS screen - most likely the driver is constantly restarting or so. When it is visible, no data shown (all fields empty).

![image](https://github.com/Louisvdw/dbus-serialbattery/assets/311609/41f12956-412a-4943-97a7-1d322048116e)

![image](https://github.com/Louisvdw/dbus-serialbattery/assets/311609/8cd607aa-b083-4715-98fd-ea5d967b5d20)

See logs bellow for details. It looks like there is `Cells populated: 0` which leads to `TypeError: '>' not supported between instances of 'float' and 'NoneType'`. Can anyone give me any ideas ?

Similar problem with zero cells I noticed with `Daly_Can`. 

I tested with `master` (commit `e3206f7`) and `dev` (commit `5e025e6`). Project didn't run with latest release since I had a python wheels issue.

Hardware: RaspberryPI zero 2w with clean VenusOS 3.14 large image, CAN Bus adapter MCP2515/TJA1050, JK BMS B2A8S20P heat can.

CAN adapter is connected according to: https://forums.raspberrypi.com/viewtopic.php?t=141052, with `dtoverlay=mcp2515-can0,oscillator=8000000,interrupt=12` in boot `config.txt`. I confirmed that my `can0` works against computer at 250kbps. I see "some" can messages from the JK bms on the computer as well in the canable `cangaroo` app (all are on the same can bus).

![image](https://github.com/Louisvdw/dbus-serialbattery/assets/311609/4acd4a42-f1bd-4373-97f6-26ac785dace3)

`candump can0`
```
  can0  2F5   [8]  83 00 9F 0F 04 AA 4A 01
  can0  2F5   [8]  83 00 9F 0F 04 AA 4A 01
  can0  4F5   [8]  DD 0C 01 DC 0C 02 55 AA
  can0  5F5   [8]  3A 01 3A 00 11 55 00 AA
  can0  2F5   [8]  83 00 9F 0F 04 AA 4A 01
  can0  2F5   [8]  83 00 9F 0F 04 AA 4A 01
  can0  2F5   [8]  83 00 9F 0F 04 AA 4A 01
[...]

PR #952, opened 2024-02-04
Bugfix 950

p0l0us:
Fixing bug https://github.com/Louisvdw/dbus-serialbattery/issues/950

The only question is, if it worked on some devices before or not ?  If the `CAN_FRAMES` change will  break other devices or not ?

Result with the fix applied:
image
image
image


mr-manuel:
Hi, thank you for contributing. Please merge with the `dev` branch and not with the `master` branch. Please leave the old `CAN_FRAMES` as comment in the code, should it break it for someone other. Actually we have no to very little feedback for the CAN drivers.

PR #960, opened 2024-02-09
Bugfix 950 dev

p0l0us:
Fixing bug https://github.com/Louisvdw/dbus-serialbattery/issues/950

The only question is, if it worked on some devices before or not ?  If the `CAN_FRAMES` change will  break other devices or not ?

Result with the fix applied:
image
image
image


mr-manuel:
Thank you!

PR #1016, opened 2024-03-21
add Seplosv3 support

marcelrv:
This PR adds support for the Seplos v3 BMS which is using the modbus protocol


mr-manuel:
Very nice thanks! We will release a new stable version at the end of March and merge that PR after.

mr-manuel (approved):
LGTM

marcelrv:
> Very nice thanks! ~We will release a new stable version at the end of March and merge that PR after.~
> 
> I add it in the next release with a note that it wasn't completely tested yet.

Great. That will also make the testing easier for others.
Here it is working fine, but indeed would need more testing esp for people with multiple Seplos batteries connected on single rs485

PR #1083, opened 2024-06-15
Bug fix: for renaming venus-data.tar.gz in post-hook.sh

juswes:
Bug fix: Renaming of the venus-data.tar.gz file to venus-data_installed.tar.gz in post-hook.sh was not working, because of wrong use of '$dir'

### DIFF
--- a/rc/post-hook.sh
+++ b/rc/post-hook.sh
@@ -26,7 +26,8 @@ bash /data/etc/dbus-serialbattery/reinstall-local.sh
 
 # rename the venus-data.tar.gz else the data is overwritten, if the USB is not removed
 for dir in /media/*; do
-    if [ -f "/media/$dir/venus-data.tar.gz" ]; then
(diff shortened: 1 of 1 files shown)


mr-manuel:
Nice catch, thanks!

mr-manuel (approved):

mr-manuel:
Are you sure? It seems strange to me, that it does not work with absolute paths.

juswes:
Yes, because the issue is that $dir already gives the absoluut path.
So in the original code:

```
for dir in /media/*; do
    if [ -f "/media/$dir/venus-data.tar.gz" ]; then
```
    
  Will be converted to:
**/media//media/mmcblk0p1**/venus-data.tar.gz

mr-manuel:
Found also another one. I pushed a commit in my repo, so it is available more faster. Thanks!

juswes:
No problem, happy to help!

PR #1087, opened 2024-06-27
added support for JK BMS with Partnumber PBxxxxx

KoljaWindeler:
This pull request adds support for the JK BMS that is used in the EEL boxes.
Their Model Number start with PBxxxxx

I've tested this implimentation for about a week and it looks like it is running fine.
It misses propper CRC checking, but I was unable to find out the CRC algorithm in use.
There is only one non-zero byte at the end of the message and it is not following any
of the common CRC algorithms. So for now I rely on a fixed length and a header detection.

Thanks, JKW

### DIFF
--- a/etc/dbus-serialbattery/battery.py
+++ b/etc/dbus-serialbattery/battery.py
@@ -59,6 +59,7 @@ class Battery(ABC):
     def __init__(self, port: str, baud: int, address: str):
         self.port: str = port
         self.baud_rate: int = baud
+        self.address: str = address
         self.role: str = "battery"
         self.type: str = "Generic"
         self.poll_interval: int = 1000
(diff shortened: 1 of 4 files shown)


mr-manuel:
Thank you for contributing to this driver. As discussed please open the PR in my repository. Alternatively I can close this PR and add it manually.

Please fill also out this table: https://github.com/mr-manuel/venus-os_dbus-serialbattery/blob/docusaurus/docs/docs/general/features.md#bms-feature-comparison

KoljaWindeler:
ha, different maintainer .. now i get it .. sorry didn't realized the change

KoljaWindeler:
i hope it still works the same as I'm not sure which version I was running on my dev environment

Issue #1062, opened 2024-05-15
Driver Fails to Start


WawaRoo:
### Describe the problem

I can't get the driver to load and connect to a USB Port.
It looks like the serial service is not starting.

### Driver version

nightly build "v1.3.20240510passthrough" 

### Venus OS device type

Cerbo GX

### Venus OS version

v3.31

### BMS type

Renogy BMS

### Cell count

4

### Battery count

1

### Connection type

Serial USB adapter to RS485

### Config file

```ini
GNU nano 4.9.3       /data/etc/dbus-serialbattery/config.ini
[DEFAULT]

; If you want to add custom values/settings, then check the values/settings you>
; and insert them below to persist future driver updates.

; Example (remove the semicolon ";" to uncomment and activate the value/setting>
; MAX_BATTERY_CHARGE_CURRENT = 50.0
; MAX_BATTERY_DISCHARGE_CURRENT = 60.0


=====================================================================================================
  GNU nano 4.9.3   /data/etc/dbus-serialbattery/config.default.ini    Modified
[DEFAULT]

; --------- Set logging level ---------
; ERROR: Only errors are logged
; WARNING: Errors and warnings are logged
; INFO: Errors, warnings and info messages are logged
; DEBUG: Errors, warnings, info and debug messages are logged
LOGGING = INFO


; --------- Battery Current limits ---------
MAX_BATTERY_CHARGE_CURRENT    = 50.0
MAX_BATTERY_DISCHARGE_CURRENT = 60.0


; --------- Cell Voltages ---------
; Description:
;     Cell min/max voltages which are used to calculate the min/max battery vol>
; Example:
<= 46.4>[DEFAULT]
```

### Relevant log output

```shell
root@einstein:~# tail -F -n 100 /data/log/serial-starter/current |tai64nlocal
2024-05-15 01:27:07.107111500 *** starting serial-starter ***
2024-05-15 01:27:07.608626500 serstart starting
2024-05-15 01:27:07.629954500 INFO: loading config file /etc/venus/serial-starte      r.conf
2024-05-15 01:27:08.458740500 INFO: loading config file /data/conf/serial-starte      r.d/dbus-serialbattery.conf
2024-05-15 01:27:10.357437500 INFO: Create daemontools service mk2-dbus.ttyS4
2024-05-15 01:27:11.949823500 INFO: Create daemontools service vedirect-interfac      e.ttyS5
[...]

PR #1124, opened 2024-09-10
robustify deployment

ppuetsch:
Sometimes it can happen that the run file or the shell scripts don't have the executable-flag set. Using this change guarantees that the executable flag is set within the deployed folder.


ppuetsch:
This might fix issues similar to this one: https://github.com/Louisvdw/dbus-serialbattery/issues/1062

mr-manuel:
I will add this lines again also in the `reinstall-local.sh`, if someone is copying files and not knowing what he is doing.

https://github.com/Louisvdw/dbus-serialbattery/blob/fd0fc1a193427df6ca0fadd141aa57826f761819/etc/dbus-serialbattery/install.sh#L247-L251

